PR #1606, opened 2014-01-14
Cron emulation

ozh:
Expand on @nicwaller's first pass to build a cron emulation and use it. Supersedes #1432


ozh:
Move all this shit to a plugin, tbh.

Issue #1719, opened 2014-06-27
Aplication out of ar


joaocaetano:
I instaled Yourls today but the admin was redirecting to http://www.salesforce.com/platform/overview/?d=70130000000liBh

The aplication is out of ar?

kamijoutouma:
MOST ANNOYING THING ever but okay 
why was this like super hard to find ...

shoeper:
This isn't super hard to find. This won't occur if one just reads how to install yourls.
If you DO NOT want to inform yourself about softwar DO NOT install it.

kamijoutouma:
I actually read the wiki and it didn't have any info about this.

Also I usually don't waste my time with closed issues. Because more then half the time it's either fixed issues or people tryna steal shit.

PR #1841, opened 2015-02-06
RFC 2606 / RFC 6761 compliance.

danielhoherd:
Changing this to be RFC 2606 / RFC 6761 compliant.  http://www.iana.org/domains/reserved

I noticed this while testing when my browser was redirected from site.com to salesforce.com.

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -1885,7 +1885,7 @@ function yourls_is_mobile_device() {
 }
 
 /**
- * Get request in YOURLS base (eg in 'http://site.com/yourls/abcd' get 'abdc')
+ * Get request in YOURLS base (eg in 'http://example.com/yourls/abcd' get 'abdc')
  *
  */
 function yourls_get_request() {
--- a/readme.html
+++ b/readme.html
@@ -337,7 +337,7 @@
 				<li>Open <code>user/config.php</code> with a raw text editor (like Notepad) and fill in the required settings</li>
 				<li>Upload the unzipped files to your domain <code>public_html</code> or <code>www</code> folder</li>
 				<li>Create a new database (see <a href="#Config" onclick="toggle(this);">Configuration</a> &ndash; you can also use an existing one)</li>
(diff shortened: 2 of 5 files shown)


LeoColomb:
:+1:

ozh:
I would prefer it to be define( 'YOURLS_SITE', 'http://your-own-domain' ); with the comment right before explaining that http://your-own-domain should be replaced with the domain -- or subdomain -- and path if applicable where the user has installed YOURLS.

The key point here being that the domain placeholder should not resolve to something actual (eg 'site.com' was, in retrospect, a bad idea that is confusing average Joe, and example.com is not better. See multiple issues about salesforce.

LeoColomb:
Should help to avoid #1719 & #1870.

ozh:
Plan:
- replace `site.com` in phpdoc comments with `example.com`, as suggested here, because it's cleaner and makes sense for devs who read the comments
- replace http://yoursite.com with http://your-own-domain-here.com/ in readme.html, user/config-sample.php and sample-public-front-page.txt to make things very self explanatory

@danielhoherd fancy updating your PR?

LeoColomb:
@ozh http://your-own-domain-here.com/ <- :laughing:

LeoColomb:
But why not `example.com` even in the config?

ozh:
Because Joe Newbie doesn't get why his install gets redirected to site.com (then salesforce.com or whatever), and will not understand more with example.com.

chtaube:
Additional idea: Do a one-time check (probably in install.php) by comparing `YOURLS_SITE` against the default value shipped in the config file. Give a friendly message to the user if he didn't change `YOURLS_SITE`.

ozh:
@chtaube Interesting idea. In the long run though that would mean two files to maintain with mostly the same content (ie default value for some var in `config-sample.php` and that same thing in `install.php/yourls_default_checks()` or whatever).

PR #1850, opened 2015-02-26
Added a find-short API action to search for the keyword of a specified l...

robvopb:
Added an API action (find-short) that finds the keyword associated with a specified long URL.  

My particular use case is to pull statistics given a long URL so I can include YOURLS stats in my aggregated metrics analysis.

### DIFF
--- a/includes/functions-api.php
+++ b/includes/functions-api.php
@@ -186,7 +186,7 @@ function yourls_api_expand( $shorturl ) {
 			'keyword'   => $keyword,
 			'shorturl'  => YOURLS_SITE . "/$keyword",
 			'longurl'   => $longurl,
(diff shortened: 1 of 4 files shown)


ozh:
Why not add that feature, indeed.

This said, there's already function `yourls_get_longurl_keywords()` to get keyword**s** (plural) associated with a long URL, it should be used. Such an API function should properly deal with the following cases:
- no return
- 1 short URL found
- several short URLs found

Also, not sure about the name "find-short". I'm not too fond of it, but at the same time I don't have a better idea...

cgmckeever:
lookup_url would also be a nice name for the endpoint

ozh:
Old PR with no feedback, closing. This would make an interesting plugin anyway.

PR #1854, opened 2015-02-28
Update functions-l10n.php

fredl99:
´yourls_load_custom_textdomain´ can't read *.mo files when ´$path´ is an absolute path on the server and the leading slash is removed. (´trim´ removes on both ends)


ozh:
Thanks !

fredl99:
I have to thank! For YOURLS itself and the quick merge of this fix.
Took me three days to track it down, because naturally I suspected my own code to be faulty. But when even the examples from the wiki were not translated, I looked a bit closer. Hmmm... ;-)

Cheers!

LeoColomb:
@fredl99 :heart_decoration: :smiley:

Issue #1303, opened 2013-04-06
URL not handled/encoded correctly


ozh:
**What steps will reproduce the problem?**
1. Enter the following long URL: http://domain.com/good space/
2. The long URL is correctly stored as http://domain.com/good%20space/
3. Now, enter the following long URL: http://domain.com/bad%20space/
4. The long URL is incorrectly stored as http://domain.com/bad%2520space/

**What is the expected output? What do you see instead?**

What I expect is that YOURLS should be smart enough to know that '%20' is a space and should be left alone.  Instead, what's happening is that YOURLS is converting the '%' to '%25' and therefore, the '%20' becomes '%2520'.

Perhaps there could be a check before cleaning up the long URL to detect '%20' or whatever else it could be (e.g. '%5C' for backspace, etc).

---

This is a **COPY** of **Issue 1303: [%20 in long URL not handled correctly](http://code.google.com/p/yourls/issues/detail?id=1303)**, filed on Google Code before the project was [moved on Github](https://github.com/ozh/google-issues-to-github).
- Submitted on 2013-02-19T06:49:18.000Z by [aylwin....@gmail.com](http://code.google.com/u/108165759648064835288/)
- Status: New

LeoColomb:
Definitely, there is a issue with encoding. May need a full rewrite.

markwaters:
Not sure if this is related , but when I try and shorten a link with a percentage sign in it , for example -
http://erinjo.xyz/content/all?q=%23Eurovision
When I visit  / expand the link , it changes to -
http://erinjo.xyz/content/all?q=#Eurovision
Which doesn't work as expected.
HTH.

adigitalife:
@markwaters I think it's tricky if the actual URL contains '%23' because it's actually the same as '#' so it gets converted automatically.  There's been a discussion that maybe YOURLS should do any kind of encoding at all.  I'm not sure what's the current status with that.

ayyoovod:
In my case, the symbol "%2F" is being replaced by "/" and "%2B" is being replaced by "+". This is also what is being formatted in (functions-formatting.php). I can't find how to change it. I tried the plugin and still not working. Any guidance is appreciated.

JeffreyDunster:
I have a similar problem with Microsoft OneDrive links. They often include %2... and %3... in their hashed IDs, which are then converted to other charters, like /, +, " ", etc.

maustyle:
hello, i still have the same issue. i have tried to install the plugin: 

[...]

PR #1859, opened 2015-03-06
Update insert.js

apermo:
Altered the function edit_link_save, Links containing encoded content like % => %25 will be further encoded to %2525 which breaks the link. As the add_link() did not contain a encoding for the url it seems save to remove it in the edit_link_save as well.

### DIFF
--- a/js/insert.js
+++ b/js/insert.js
@@ -133,7 +133,7 @@ function edit_link_hide(id) {
 // Save edition of a link
 function edit_link_save(id) {
 	add_loading("#edit-close-" + id);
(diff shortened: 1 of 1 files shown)


LeoColomb:
X-Ref: #1303

ozh:
*(no, I didn't completely ignored that PR :)*

I'm not sure this works any better, or maybe I didn't get what it's supposed to improve.
Example:
`https://www.google.com/search?q=url+with+%2525` -> things are not improved between `master` and your PR

Am I getting something wrong?

ozh:
Closing because no feedback.

PR #1860, opened 2015-03-06
added index page

frog23:
I added a index.php page to quickly create and resolve short URLs. It avoids the 403/404 error when the base URL of yourls is called. 
It renders two different forms, one for resolving a short url, and one for creating a new one (similar to the form on the admin page, but when the submit button is pressed, it forwards the content just like the bookmarklet does).
I create this page, just to "scratch my own itch" and maybe this is useful to others as well. However I can understand that some people might see no additional benefit from such a page (after all, all it does, copy the functionality of the bookmarklet and provide a text field for something you can just as well write in the URL bar of your browser). But the way I see it, this can not be done with an plugin, since the call for the base URL would never reach yourls (please correct me, if I am mistaken about this).
Alternatively this page could be renamed, and users would have to change it's name to index.php to enable it. Or it could just be documented in the wiki as is the case with [RSS.php](https://github.com/YOURLS/YOURLS/wiki/RSS.php). I will leave this decision to you.
Cheers, Frog23

### DIFF
--- a/index.php
+++ b/index.php
@@ -1,0 +1,46 @@
+<?php
+require_once( dirname(__FILE__).'/includes/load-yourls.php' );
+
+yourls_html_head();
+yourls_html_logo();
+?>
+	<ul id="admin_menu">
+		<li>Go to the <a href="<?php echo yourls_admin_url('index.php') ?>">Admin Interface</a></li>
+	</ul>
+	<div class="sub_wrap">
+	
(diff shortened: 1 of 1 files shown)


ozh:
This cannot go in, as it would break numerous installs, on yourls.org for starters. The vast majority of YOURLS users want a private URL shortener for their own needs, and as such their index.php is purposely distinct from YOURLS.

PR #1862, opened 2015-03-08
Correction in yourls_die(): 'html_head' should be 'html_footer'

BstName:
Now checking for html_footer action before the call to yourls_html_footer() (code was checking for html_head action instead, which is wrong).


LeoColomb:
Good catch! :mag: 
Thanks a lot! :+1:

BstName:
You're welcome - and thanks for YOURLS!

PR #1875, opened 2015-03-20
Sanitize only after applied filters

freimanas:
Keyword may be changed in filter, so makes sense to sanitize after that.


ozh:
This change could potentially break any function or plugin using filters `custom_keyword` or `random_keyword` if the escaping changes the expected output. Anyway, there's also a filter in `yourls_escape()` that you can use to torture things more.

PR #1882, opened 2015-04-10
Smarter domain convention in examples and source

ozh:
In config, readmes, editable files: the generic YOURLS install will be http://your-own-domain-here.com/ so that noob users get an obvious warning when they misconfigure something

In source comments, presumably read by hackers who have a clue, the generic YOURLS install will be http://sho.rt/. The generic long URL can be http://example.com/something-whatever

Closes #1841


ozh:
(leaving that PR open while we make sure there's no other place to edit that way)

PR #1921, opened 2015-06-26
add HTML5 tag and ARIA landmark

audreymaniez:
I added HTML5 tag (`header nav footer main`) and ARIA landmark (`navigation main banner contentinfo`) on templates.
[See WAI-ARIA recommendation ](http://www.w3.org/TR/wai-aria/roles#landmark_roles_header) 

No design impact (no need to modify CSS files)

And that's really good for screen reader users :-)

If interested, i can go further and I can work on making it all accessible, as I done for our society installation.

### DIFF
--- a/admin/tools.php
+++ b/admin/tools.php
@@ -8,8 +8,9 @@ yourls_html_logo();
 yourls_html_menu();
 ?>
 
-	<div class="sub_wrap">
-	
+	<div class="sub_wrap">	
+	<main role="main">
+
 	<h2><?php yourls_e( 'Bookmarklets' ); ?></h2>
 	
 		<p><?php yourls_e( 'YOURLS comes with handy <span>bookmarklets</span> for easier link shortening and sharing.' ); ?></p>
@@ -329,7 +330,8 @@ $signature = md5( $timestamp . '<?php echo yourls_auth_signature(); ?>' );
 	</ul>
 	
 	<p><?php yourls_se( 'See the <a href="%s">API documentation</a> for more', YOURLS_SITE . '/readme.html#API' ); ?></p>
-
+		
+	</main>	
 	</div>
 
 	<?php } // end is private ?>
--- a/includes/functions-html.php
+++ b/includes/functions-html.php
@@ -7,10 +7,12 @@
 function yourls_html_logo() {
 	yourls_do_action( 'pre_html_logo' );
 	?>
+	<header role="banner">
 	<h1>
 		<a href="<?php echo yourls_admin_url( 'index.php' ) ?>" title="YOURLS"><span>YOURLS</span>: <span>Y</span>our <span>O</span>wn <span>URL</span> <span>S</span>hortener<br/>
 		<img src="<?php yourls_site_url(); ?>/images/yourls-logo.png" alt="YOURLS" title="YOURLS" border="0" style="border: 0px;" /></a>
 	</h1>
+	</header>
 	<?php
 	yourls_do_action( 'html_logo' );
 }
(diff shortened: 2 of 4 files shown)


LeoColomb:
Big :+1: for the goal! :smile: 
That said, I'm not a big fan of all these new HTML elements, and ARIA conventions are focused on roles, not HTML5 elements. 
See my inline comments. :arrow_down:

admin/tools.php
```diff
@@ -8,8 +8,9 @@
 yourls_html_menu();
 ?>
 
-	<div class="sub_wrap">
-	
+	<div class="sub_wrap">	
+	<main role="main">
```
LeoColomb: For example, here, I think `main` and `div.sub_wrap` should be merged.

includes/functions-html.php
```diff
@@ -140,14 +142,14 @@ function yourls_html_footer() {
 	
 	$num_queries = sprintf( yourls_n( '1 query', '%s queries', $ydb->num_queries ), $ydb->num_queries );
 	?>
-	</div> <?php // wrap ?>
-	<div id="footer"><p>
+	</div><?php // wrap ?>
+	<div id="footer"><footer role="contentinfo"><p>
```
LeoColomb: Same here, with `footer` and `div#footer`.

includes/functions-html.php
```diff
@@ -757,7 +760,7 @@ function yourls_html_menu() {
 	$admin_sublinks = yourls_apply_filter( 'admin_sublinks', $admin_sublinks );
 	
 	// Now output menu
-	echo '<ul id="admin_menu">'."\n";
+	echo '<nav role="navigation"><ul id="admin_menu">'."\n";
```
LeoColomb: Again with `nav` and `ul#admin_menu`.

LeoColomb:
> And that's really good for screen reader users

And just for the joke, I'm quite curious of who use YOURLS on a screen reader! :laughing: :computer:
At the end, thanks a lot for the contribution.

audreymaniez:
Hey :-)
One of our colleague use Yourls with a screen reader. By screen reader users, I mean visual deficient user.

admin/tools.php
audreymaniez: Ok for this one
I wanted to keep the div that has a position in YOURLS css, because of HTML5 support in old browser. But it's better if you merge the two.
But you can add support with a js&css shiv : https://github.com/afarkas/html5shiv

includes/functions-html.php
audreymaniez: Ok for footer

includes/functions-html.php
audreymaniez: Nop :-)
Because `nav` doesn't have the same HTML role definition as an `ul` so keep `<nav role="navigation"><ul id="admin_menu">`

LeoColomb:
> By screen reader users, I mean visual deficient user.

My bad, you're totally right. :smiley:

includes/functions-html.php
LeoColomb: In term of HTML element, you're right. But, as I said in my general comment, I don't think you _have_ to use HTML5 elements to add ARIA roles. 
Am I wrong?

audreymaniez: Well, HTML elements are semantic :-) 
`footer` means that it's a footer. A div means nothing. You can see it is a footer because ... well you can see. Someone who can't see, doesn't know it is a footer. So landamark say its a block with content info, and html5 `footer` say its a footer :-D 
Do i convince you ? 

Even if HTML5 elements are not well supported in screen reader today, they will, and it's not that bad to use the new HTML5, the new W3C Recommendation ;-)

Semantic is good for the web :-)

ozh:
Any CSS fix to add?

@LeoColomb you merge this one when you're OK with it

audreymaniez:
No CSS fix, in my local testing, no impact with my commit.

includes/functions-html.php
LeoColomb: > Do i convince you ?

I'm already convinced by HTML5, but just disappointed by so "huge" change on HTML4 DOM structure which can raise incompatibility issues for some old users.
I'm just saying YES for ARIA roles, but NOT SURE for some alone HTML5 elements lost in an older structure, which are two separate and different changes.

LeoColomb:
@audreymaniez Nevertheless, I'm globally OK for your pretty cool contribution, can you just merge elements when possible (keep ids/classes, for CSS!)? :smile: 
Thanks again! :+1:

PR #1935, opened 2015-07-29
use correct http status code for rate limiting

cemeyer2:
403 is not the ideal status code to return when being rate limited/throttled. 429 is more semantically correct


ozh:
Agreed. I'm postponing the merging of this PR because it introduces a new translatable string, but that is :+1:

LeoColomb:
Big :+1:

ozh:
(nice job coveralls... 18 months late and analysed the wrong project :) )

LeoColomb:
I think this can be merged, even if it introduces a new translatable string.
Actually, statuses for API should not be translated.

PR #1937, opened 2015-07-29
Update forbidden to locked on rtn code

williamrhancock:
Update YOURLS_FLOOD_DELAY return HTTP code to proper status of locked from forbidden.


ozh:
I'd rather send a 419, see #1935

PR #1942, opened 2015-08-06
Fix CSS on the admin page

tomslominski:
Currently, the admin pages of 2.0 link to yourls.css, but the file is actually called yourls.min.css. This corrects that and enables the CSS to work.


LeoColomb:
Yep.

But will move in the future.

PR #1988, opened 2015-11-27
Make password auto encryption work when password contains slash(es)

### LINKED ISSUE #1987: Auto encrypting passwords doesn't work if password contains forward slashes
If the password contains one or more forward slashes, auto encryption doesn't work and raises a `'preg_replace problem'`.

The culprit is the prog_quote call in line 173 of includes/functions-auth.php. Should be `preg_quote( $password, '/' )` instead of `preg_quote( $password, '-' )`

Thanks,

Bernie

berndblume:
Fix #1987

### DIFF
--- a/includes/functions-auth.php
+++ b/includes/functions-auth.php
@@ -170,7 +170,7 @@ function yourls_hash_passwords_now( $config_file ) {
 			// PHP would interpret $ as a variable, so replace it in storage.
 			$hash = str_replace( '$', '!', $hash );
 			$quotes = "'" . '"';
-			$pattern = "/[$quotes]${user}[$quotes]\s*=>\s*[$quotes]" . preg_quote( $password, '-' ) . "[$quotes]/";
+			$pattern = "/[$quotes]${user}[$quotes]\s*=>\s*[$quotes]" . preg_quote( $password, '/' ) . "[$quotes]/";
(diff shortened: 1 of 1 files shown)


berndblume:
Travis-CI doesn't run through because composer is outdated.

LeoColomb:
Looks good! Thanks :smile:

ozh:
If I'm correct, this doesn't work with a password like `'lol .\+*?[^]$(){}=!<>|:-/' . "'" . '"'` (ie containing all preg special chars)

berndblume:
It should work with that password because prog_quote does quote ALL special prog chars: [link to php docs] (http://php.net/manual/en/function.preg-quote.php). I haven't tested that particular password ;-), but it should work.

The thing is: for some reason, the second argument ($delimiter) to prog_qoute was passed as '-', while in reality we're using the standard '/' as the delimiter. After my change, passing '/' as the delimiter rather than '-', it works with passwords that contain slashes as well as with passwords that contain dashes. For all the other special prog characters, prog_quotes them automatically independent of the $delimiter argument.

In essence: everything should be fine by now.

ozh:
My comment was based on actual tests with the unit test suite. It doesn't work. Not sure why, didn't dig deep enough, but it doesn't work.

ozh:
I stand corrected -- my test was flawed, your PR indeed fixes the forward slash problem. Much thanks!

PR #1991, opened 2015-11-29
added 400+ bad words

paulmadore:
Found this list here: http://urbanoalvarez.es/blog/2008/04/04/bad-words-list/

I think it creates further protection from abuse. Since it's just the config sample, it's not as if the user has to use it, but the novice may have a lessened chance of their YOURLS being abused.


ozh:
This is definitely plugin territory :
- bad words for you may not be bad words for others
- YOURLS is used by people from all over the world speaking multiple languages, having 400+ English words is useless to many
- if anything, the goal would be to take settings **out** of the config files, not add to id

A much better solution would be to make a plugin, with a configurable list of forbidden words

paulmadore:
>> A much better solution would be to make a plugin, with a configurable list of forbidden words

Okay, perhaps I'll do that in the next week or so. Less busy than usual. Thanks for responding so quick.

PR #2012, opened 2015-12-29
Using HTTP_HOST as default installation URL

liruqi:
Support multiple domains, simplify configuration & deployment


ozh:
Nah. Much more confusing for the average Joe. The whole point of the line `define( 'YOURLS_SITE', 'http://your-own-domain-here.com' )` is to have noob easily understand what they have to edit, especially in case when YOURLS is in a directory and not off the domain root. Having a cryptic variable here will not help.

PR #2014, opened 2016-01-02
Adding more reserved URL keywords (Bad words list)

SeanZoR:
According to http://urbanoalvarez.es/blog/2008/04/04/bad-words-list/


dgw:
@ozh will not merge this. See #1991.

LeoColomb:
Thanks @SeanZoR!
But @dgw is right... See #1991.

PR #2038, opened 2016-02-15
Add an option to set temporary redirect code

SnoFox:
I've currently just manually modified the file doing this, but as I'm setting this up in Docker, I'd like to use your repo as the source (rather than my own)

Hopefully this is acceptable :]


ozh:
You're doing it wrong, there is a plugin api just for this kind of behavior. I'm sure there is even a plugin for this already.

As a general rule, there should be LESS options rather than more.

PR #2043, opened 2016-03-01
Fix f. missing chart bar add (+all) icon

BstName:

When allowing for multiple short URLs for a same long URL (bitly behavior [behaviour]), the chart bar adding symbol linking to YOURLS_SITE/keyword+all does not show up on infos pages.
This issue exists at least since YOURLS 1.7.0.

To reproduce this bug e.g. with a fresh install, first set 

define( 'YOURLS_UNIQUE_URLS', false );

in YOURLS/user/config.php. Then, in the admin panel, create two short URLs pointing both to a same long URL, e.g. to http://yourls.org. Let us assume that the short URLs are /1 and /2. Now visit the infos page for one of the short URLs, e.g. by visiting YOURLS_SITE/1+ in your browser (or by clicking the chart icon in the respective row for that short URL in the admin panel's table).

Sadly, no clickable chart bar adding symbol linking to YOURLS_SITE/1+all (depicting chart_bar_add.png) shows up on the infos page.

Fix (in yourls-infos.php): 

The canonical way to fix this is to just move the line 

`$keyword_list = yourls_get_longurl_keywords( $longurl );`

out of the if-statement testing $aggregate. That way, the `$keyword_list` is set correctly regardless of the value of $aggregate. This is needed later on:

```
if( isset( $keyword_list ) && count( $keyword_list ) > 1 )
    {code echoing the +all-link, depicting chart_bar_add.png}

```
Improvement:

And since `$keyword_list` is only needed when duplicate long URLs are allowed, we can avoid the DB access involved with a call to ` yourls_get_longurl_keywords()` in the case that duplicate long URLs are not allowed:

```
íf( yourls_allow_duplicate_longurls() ) 
    $keyword_list = yourls_get_longurl_keywords( $longurl ); 

```


ozh:
Took me a while to figure out what wasn't working with your code, because, yeah, your code doesn't work. `íf` instead of `if`, I really wondered how you managed to get that in ...

Good catch nonetheless, you should fix it so I can merge the PR

BstName:
Not sure what happens here: strangely enough, Travis CI build now fails (see details) though the typo is corrected (and as you already mentioned yourself, it passed WITH the typo before). I then tried a new identical PR after pushing a fresh fetch from YOURLS/master to my fork, from a new branch (#2049). Same result. I close this PR, but keep the fresh one (#2049) open for a while, in case you would like to investigate what exactly happens here.

Pitch black magic going one here?

BstName:
OK, meanwhile #2049 has been merged by @LeoColomb.
Thanks to @ozh for looking into this and for correcting the typo.

PR #2049, opened 2016-03-04
Fix f. missing chart bar add (+all) icon

BstName:

When allowing for multiple short URLs for a same long URL (bitly behavior [behaviour]), the chart bar adding symbol linking to YOURLS_SITE/keyword+all does not show up on infos pages.
This issue exists at least since YOURLS 1.7.0.

To reproduce this bug e.g. with a fresh install, first set 

define( 'YOURLS_UNIQUE_URLS', false );

in YOURLS/user/config.php. Then, in the admin panel, create two short URLs pointing both to a same long URL, e.g. to http://yourls.org. Let us assume that the short URLs are /1 and /2. Now visit the infos page for one of the short URLs, e.g. by visiting YOURLS_SITE/1+ in your browser (or by clicking the chart icon in the respective row for that short URL in the admin panel's table).

Sadly, no clickable chart bar adding symbol linking to YOURLS_SITE/1+all (depicting chart_bar_add.png) shows up on the infos page.

Fix (in yourls-infos.php): 

The canonical way to fix this is to just move the line 

`$keyword_list = yourls_get_longurl_keywords( $longurl );`

out of the if-statement testing $aggregate. That way, the `$keyword_list` is set correctly regardless of the value of $aggregate. This is needed later on:

```
if( isset( $keyword_list ) && count( $keyword_list ) > 1 )
    {code echoing the +all-link, depicting chart_bar_add.png}

```
And since `$keyword_list` is only needed when duplicate long URLs are allowed, we can avoid the DB access involved with a call to ` yourls_get_longurl_keywords()` in the case that duplicate long URLs are not allowed:

```
íf( yourls_allow_duplicate_longurls() ) 
    $keyword_list = yourls_get_longurl_keywords( $longurl ); 

```


LeoColomb:
OK, thanks! (the Travis error is not introduced by this change)

BstName:
OK, thanks a lot :+1:

PR #2051, opened 2016-03-05
Fixing markup and comment

BstName:
Minor fixes in yourls-infos.php: 
1. Removes an unclosed and superfluous `<strong>` tag in the line reading e.g.  '43 hits on January 1, 1970'. This caused wrong HTML and therefore most browsers rendered *all* dates in the list appearing on click of 'Click for more details' with bold font-weight (i.e., not only those of the best year / month / day as intended).
2. Took the liberty of also changing a closing comment token from `/**/` to the more common `**/` (in the comment titled 'I can haz debug data'). While this was perfectly valid code, it confused the syntax highlighter of github (all past the `/**/` appeared in a frightening red).

### DIFF
--- a/yourls-infos.php
+++ b/yourls-infos.php
@@ -198,7 +198,7 @@ if( yourls_do_log_redirect() ) {
 	echo "last_24h: "; print_r( $last_24h );
 	echo "countries: "; print_r( $countries );
 	die();
-	/**/
+	**/
 
 }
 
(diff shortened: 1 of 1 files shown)


LeoColomb:
Good catches! Thanks! :+1:

BstName:
WOW, faster than light, thanks :smiley:

PR #2054, opened 2016-03-09
Go to First Page -> Go to Last Page

BstName:
Fixes a minor glitch in `yourls_html_tfooter()` in functions-html.php:

In the pagination section of the URL table, the 'Go to Last Page'-link has a `title` attribute 'Go to First Page' (one can notice that by hovering it). Should of course be titled 'Go to Last Page'.


LeoColomb:
Indeed! :smile: 
Thanks!

BstName:
You are welcome, thanks for merging.

PR #2055, opened 2016-03-10
Updating total clicks on delete

BstName:

Let us assume we have a keyword with a non-zero number of clicks. 
If the respective row of the URL/keyword table in the admin panel is deleted (via the 'Delete' button in the 'Actions' column), the total number of URLs (or links, respectively) is properly updated (i.e., decremented) in the summary on the top of the page, but the total number of clicks (i.e., the '{x}' in '{x} clicks, and counting!') remains unchanged.

This bug exists at least since v1.7.0.

To reproduce the bug, create a new keyword (say, a new custom keyword '4test') and give it a few clicks (by clicking the link in the 'Short URL' column of the respective new row for '4test', or by visiting 'http://{your-yourls-domain.tld}/4test' a few times). Now, after a refresh of the admin page, memorize or write down the number of total clicks as stated in the summary line '{x} clicks, and counting!'. Delete the row for '4test' by clicking the 'Delete' button in the 'Actions' row. Sadly, '{x}' is not decreased, but shows the same value as before.

Fix: First granted the `<strong>` tag containing the number of total clicks its own id (in admin/index.php) - the element should be prominent enough to deserve it. Added a function `decrease_total_clicks( id )` in js/insert.js which decreases that element by the value contained in the element `$('#clicks-' + id)` (i.e. by the value picked from the 'Clicks' column of the respective deleted row). Now called this method at the appropriate place in function `remove_link(id)`, thereby passing on `id` as parameter.

Please advise if you have different preferences for naming and/or coding style - PR will be updated accordingly.

### DIFF
--- a/admin/index.php
+++ b/admin/index.php
@@ -286,7 +286,7 @@ if ( !$is_bookmark ) { ?>
 			echo ", " . sprintf( yourls_n( 'counting <strong>1</strong> click', 'counting <strong>%s</strong> clicks', $total_items_clicks ), yourls_number_format_i18n( $total_items_clicks ) );
 	?>.</p>
 <?php } ?>
-<p><?php printf( yourls__( 'Overall, tracking <strong class="increment">%1$s</strong> links, <strong id="total_clicks">%2$s</strong> clicks, and counting!' ), yourls_number_format_i18n( $total_urls ), yourls_number_format_i18n( $total_clicks ) ); ?></p>
+<p id="overall_tracking"><?php printf( yourls__( 'Overall, tracking <strong class="increment">%1$s</strong> links, <strong>%2$s</strong> clicks, and counting!' ), yourls_number_format_i18n( $total_urls ), yourls_number_format_i18n( $total_clicks ) ); ?></p>
 <?php
 
 yourls_do_action( 'admin_page_before_form' );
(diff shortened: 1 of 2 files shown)


ozh:
Nice catch indeed. Ideally this should be done without changing the HTML (the new `id` tag) so it could be merged without requiring all translation files to be updated as well, see https://github.com/YOURLS/YOURLS.pot/blob/master/YOURLS.pot#L266-L267

BstName:
Thanks to @dgw and @ozh for your feedback!

@ozh: Oh, I see. 

It think it is non-trivial to do this without any change in admin/index.php since one would have to account for a possible introduction of more elements with `class="increment"` and/or of more `<strong`> tags by the two actions surrounding the code in question (actions 'admin_page_before_content' and 'admin_page_before_form'). 

But there might be an easier solution: how about giving the paragraph element (`<p></p>`) enclosing the text 'Overall, tracking ...' an id? Since this does not affect the string passed to `yourls__`, updating the .pot / recompiling the .po´s should not be necessary (is this correct, or am I missing something in the process chain used here?) 

If this is correct, one can use this id as a convenient anchor for selection, e.g. like so (in function `decrease_total_clicks( id )`:

```
var total_clicks_selector = "#overall_tracking strong:nth-child(2)";
$(total_clicks_selector).html( parseInt( $(total_clicks_selector).html() ) - parseInt( $('#clicks-' + id).html() ) );

```
(thereby assuming a newly given id="overall_tracking" for the said paragraph).

What do you think, should I update the PR following this idea?

BstName:
Well, I just updated the PR - please revise if there is still something that must be changed. Please notice that I will be on travel from Tuesday afternoon to Sunday evening, so there will probably not be any responses from me during that time. Wishing you a great weekend :smiley:

ozh:
Thanks!

BstName:
Great, thanks for merging!

PR #2059, opened 2016-03-16
Update readme.html

matecode:
added aYourls to Android apps


ozh:
Thanks! Now pimped on http://yourls.org/#More as well

PR #2063, opened 2016-03-19
Make YOURLS_SITE able to be different from installpath 'YOURLS_URL'

damadmai:
This makes it possible that YOURLS can be installed in a subdirectory of your web root, so that your web root is not cluttered up with the files of YOURLS.

This introduces a new setting YOURLS_URL where you specify the URL of your YOURLS installation so that the admin interface can work:
`define( 'YOURLS_URL', 'https://example.com/yourls' );`
This is not your short link base URL. / It can differ from it.
This setting can be ommitted if YOURLS is installed in the short link base web root as it would be normally.

Fixed some whitespace issues in ./user/config-sample.php

### DIFF
--- a/admin/tools.php
+++ b/admin/tools.php
@@ -309,7 +309,7 @@ TUMBLR;
 		<ul>
 			<li><h3><?php yourls_e( 'Usage of the signature token' ); ?></h3>
 			<p><?php yourls_e( 'Simply use parameter <tt>signature</tt> in your API requests. Example:' ); ?></p>
-			<p><code><?php echo YOURLS_SITE; ?>/yourls-api.php?signature=<?php echo yourls_auth_signature(); ?>&action=...</code></p>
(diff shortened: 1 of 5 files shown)


ozh:
As stated in several issues & PR : more options = more confusion for the average user = more support requests. If anything, we'll have *less* options.

The feature you describe is indeed interesting, but it would be much better as a plugin.

damadmai:
I looked at the implementation of plugins, but i see no possibility to implement this feature as a plugin.
The Admin Control Panel for enabling plugins is not accessible if the hard coded URLs for it are not changeable in the config!
I understand your point and already thought of it as I introduced the new setting only in  ./includes/load-yourls.php as a hidden one and not in ./user/config-sample.php.
Btw. do you know why the Travis CI build fails?

ozh:
Then let's add what's needed to make it possible to implement this feature as a plugin :) 
Function `yourls_site_url()` could be used more, since it returns a filtered result, instead of using const `YOURLS_SITE`

Issue #2067, opened 2016-03-22
Keep 5 versions of JQuery


Preovaleo:
Hello,

Is it relative to keep 5 versions of JQuery?

regards

LeoColomb:
Not sure to understand your question. Could you explain, please?

Preovaleo:
autocorrect :-1: relative => relevant

is this relevant to keep 5 version of jquery ? from 1.3 to 1.9

LeoColomb:
Absolutely not, it's not relevant. Why do you ask? What is in your mind?

Preovaleo:
So could we remove 4 of them ?

LeoColomb:
Aaah right, got it! Well, need @ozh review.

ozh:
Hmm, yeah, indeed. I've been probably a bit too conservative here :) Keeping latest jQuery only, renamed `jquery.js`, should be enough, especially now that jquery dev is much slower than it used to be a couple years ago

PR #2070, opened 2016-03-23
Reduce the number of jquery and bump

Preovaleo:
solve  #2067


ozh:
Does jQuery 2.x break things ?

Preovaleo:
I don't know but yes it could be a good idee

Preovaleo:
yes it's working

.gitignore
```diff
@@ -50,3 +50,6 @@ Desktop.ini
 
 # Mac crap
 .DS_Store
+
+# NetBeans files
+/nbproject/
```
GaryJones: This IDE / OS specific stuff should be in your own global gitignore, not at the project level.

PR #2094, opened 2016-05-14
Access-Control-Allow-Origin for ajax calls on CNAMES

vincentgiraud:
Add CORS header fix for same domain, else ajax calls for actions will break if using DNS CNAMES.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -50,3 +50,6 @@ Desktop.ini
 
 # Mac crap
 .DS_Store
+
+# IDEs
+.idea
--- a/includes/functions-html.php
+++ b/includes/functions-html.php
@@ -908,6 +908,7 @@ function yourls_content_type_header( $type ) {
 	if( !headers_sent() ) {
 		$charset = yourls_apply_filter( 'content_type_header_charset', 'utf-8' );
 		header( "Content-Type: $type; charset=$charset" );
+		header("Access-Control-Allow-Origin: "+YOURLS_SITE);
 		return true;
 	}
 	return false;


ozh:
Can you describe a scenario where ajax calls fail? (not used to using DNS CNAMES)

vincentgiraud:
Setting: a sub domain aaa.xxx.tld with CNAME bbb.ccc.tld behind SSL
Scenario: after loging in into YOURLS, I click the edit button on the right for an existing shortURL row 
Behavior: endless spin (the ajax is stopped because of CORS)
"XMLHttpRequest cannot load https://bbb.ccc.tld/admin/admin-ajax.php?action=edit_display&keyword=1&nonce=xxxxxxxxxx&id=wwwwwwwwww. No 'Access-Control-Allow-Origin' header is present on the requested resource. Origin 'http://bbb.ccc.tld' is therefore not allowed access."
Solution proposed: Access-Control-Allow-Origin: https://bbb.ccc.tld

ozh:
I see

Digging further I see a few problems that could raise:
- `YOURLS_SITE` is not necessarily a domain, it can be a path (eg `http://mydomain.com/yourls`) so I'm not sure if that would affect some setups
- There's a plugin called "allow aliases" which I think works around your edge case, that would be probably broken with this new
- Several folks have developed plugin or tricks to handle multiple domains with one YOURLS install, and that could break or not with your code

All in all, I don't want to introduce code supposed to fix an edge case and potentially breaking other edge cases.

Your perfectly valid need for a fix has a much simpler solution: there's a [hook to inject custom headers](https://github.com/YOURLS/YOURLS/blob/master/includes/functions-html.php#L66), so make a plugin to tap into it and send what you need.

Simple plugin:
```php
<?php
/*
all the plugin header stuff
*/

yourls_add_action( 'admin_headers', 'giraud_fix_cors' );

function giraud_fix_cors() {
    header("Access-Control-Allow-Origin: " . YOURLS_SITE); // BTW, note: . , not +
}
```

vincentgiraud:
Following your advise, I've created a plugin for whoever may be in this case, and url parsed YOURLS_SITE to ignore a potential path:
https://github.com/TEODE/yourls-access-control-allow-origin

ozh:
I don't think your plugin works as intended, because of `"Access-Control-Allow-Origin: "+$fqdn`, which should be `"Access-Control-Allow-Origin: ".$fqdn`

See https://3v4l.org/27Q1J

vincentgiraud:
Thanks for picking that up, I obviously overlooked some testing here :) Updated!

ozh:
Added to the plugin list! :balloon:

Issue #1898, opened 2015-05-04
More validation of plugin files


ozh:
You can activate a file that's not a plugin at all, see #1897
- this file appears with `(no info)` as name, author, version and description: I think at least a name should be mandatory
- `(no info)` as plugin name and plugin author are absurdly linked to `http://sho.rt/admin/(no%20info)`
- is there any way to check that the file is at least some PHP ?

dgw:
What about checking that the file contains `<?php`? Not foolproof, certainly, but better than no check at all. HTML should fail that test because `<` needs to be escaped.

vaughany:
Could you require that plugin authors provide author name, plugin name, plugin version and plugin release date inside a function which you can call from the core code? A failure to find the function, or if null/empty return would prevent the plugin being made live.

ozh:
Requiring all the info won't help making sure the plugin is valid, requiring a release date would also break 100% of existing plugins which we won't do

Ideally we need a sandboxed activation process, like it's done in WordPress, but I haven't butchered and dissected their way of doing it yet :)

fredl99:
> (no info) as plugin name and plugin author are absurdly linked to http://sho.rt/admin/(no%20info) 

If this behaviour is reliable and reproducable then why not ship a prepared (static) page with that name? 

It could contain some information that there's something wrong with that file, maybe a hint to upgrade or some other notes. If it's not static the page could also tell the plugin's filename and/or a link to deactivate it.

Not a solution but at least some use for that link. Just an idea...

peterberbec:
Quick and dirty <php? test

ozh:
We should maybe have a look at https://github.com/Corveda/PHPSandbox. Then, if not trivial to use and 100% foolproof, this issue will be closed

ozh:
During all these years, I've never read from someone troubled by this lack of verification. Closing.

PR #2105, opened 2016-06-07
#1898 - Plugin validation I

peterberbec:
https://github.com/YOURLS/YOURLS/issues/1898

let's verify the file at least starts with <?php

### DIFF
--- a/includes/functions-plugins.php
+++ b/includes/functions-plugins.php
@@ -440,7 +440,7 @@ function yourls_load_plugins() {
  * @return bool
  */
 function yourls_validate_plugin_file( $file ) {
-	$firstchars="";
+	$firstchars="";					// initialize, if we file isn't readable, an error would be thrown
 	if ( is_readable( $file ) ) {			// if the plugin file is readable
 		$fp = fopen( $file, 'r' );		// open the file for reading
 		$firstchars = fread( $fp, 10 );		// read the first 10 characters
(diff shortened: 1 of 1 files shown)


ozh:
Sane idea, but very confusing code IMHO:
```
	if ( ... ) {
		if ( ... )
		return false;
	}
	else return false;
	return true;
```

peterberbec:
is it the lack of comments, or the nested ifs? let me think on it for a while, I might be able to get another way.

peterberbec:
This reads much better.

If this gets in, are the tests of the filename still needed? We have already opened the file and proved it starts with `"<?php"`. Could
`		false !== strpos( $file, '..' )		// check filename for sanity I
		OR
		false !== strpos( $file, './' )		// check filename for sanity I
		OR`
be removed?

ozh:
Back on this PR after so long

I'm dropping the idea because depending on how the PHP plugin file is saved (UTF8 with BOM or without BOM), the first 5 **or 8** bytes are needed to fetch the opening `<?php`. (For most folks, fetching first 5 bytes [is OK](https://3v4l.org/RISfr) and the file begins with `<?php` at position 0, but for some folks we'd need 8 bytes and the file would begin with `<?php` at position **3**.)

So, I don't think this is worth the hassle and potential troubles. I'm all ears if someone has a *simple* way of checking this ("simple", as in "will work in 100% cases"), otherwise let's just assume that folks creating a file named `plugin.php` and not starting it with `<?php` just deserve what they'll get :)

PR #2144, opened 2016-09-16
unit test with php 7.1

Ramoonus:


ozh:
Hmm. Nasty fails.

ozh:
The problem is that PHPUnit versions on Travis differ and they're not using the same syntax. `PHPUnit_Framework_TestCase` is now `PHPUnit\Framework\TestCase` in PHPUnit 6.

PHPUnit 5.7 has a forward compatibility layer, meaning you can already use PHPUnit\Framework\TestCase with that version, there is no backward compatibility layer in PHPUnit 6 that allows you to use PHPUnit_Framework_TestCase 

As of writing, on Travis:

| PHP version | PHPUnit version
| -| -
| 5.3, 5.4, 5.5, 5.6 | PHPUnit 4.5.0
| 7.0, 7.1 | PHPUnit 6.0.9

For the record, PHPUnit 4.8 does not support PHP 7, while 5.1 requires PHP 5.6+

At the moment it seems most folks are just skipping PHPUnit 6 and forcing use of 5.7

Possible workarounds:
1) force PHPUnit 5.7 on PHP7+, like [they do for WP](https://github.com/WordPress/wordpress-develop/blob/master/.travis.yml#L66-L87) or [Symfony](https://github.com/symfony/symfony/blob/master/src/Symfony/Bridge/PhpUnit/bin/simple-phpunit#L19) (they also specify [PHPunit version conflicts](https://github.com/symfony/symfony/blob/master/composer.json#L107), not sure what the point is)
2) Update composer.json to restrict PHPunit versions, like [Codeception](https://github.com/Codeception/Codeception/blob/2.2/composer.json#L21)

ozh:
For the record, since I cannot get the correct PHPunit version to run, it's [supposed to be like this](https://docs.travis-ci.com/user/languages/php/#Default-Test-Script):

> Travis CI looks for phpunit in the following order (as Composer does) and uses the first one found.
>
> -  $COMPOSER_BIN_DIR/phpunit
> -   phpunit found in the directory specified by bin-dir in composer.json
> -   vendor/bin/phpunit
> -   phpunit, which is found on $PATH (typically one that is pre-packaged with the PHP runtime)

ozh:
For fuck sake, this one is supposed to work [as per the doc](https://phpunit.de/manual/current/en/writing-tests-for-phpunit.html#writing-tests-for-phpunit.exceptions.examples.ErrorTest.php)

ozh:
Almost there.

The [problem now](https://travis-ci.org/YOURLS/YOURLS/jobs/222522714#L308) is because in PHP 7.1, former warnings when passing too few arguments to a function got raised to an ArgumentCountError exception. Not sure how to deal with this differently between PHP 5.3-7 and PHP 7.1 ...

LeoColomb:
Wow, so complicated! But interesting and complete study. 👌 

> Not sure how to deal with this differently between PHP 5.3-7 and PHP 7.1

Actually, warnings should have been removed, for code quality and compilation reliability.

PR #2158, opened 2016-10-09
Added a "delete" API action to delete a specific shorturl

werpat:
Added a "delete" API action to delete a specific given shorturl. 

I use yourls in combination with owncloud to shorten the owncloud share-links and i would like to remove shorturls when the corresponding owncloud-url is removed.

### DIFF
--- a/includes/functions-api.php
+++ b/includes/functions-api.php
@@ -83,6 +83,18 @@ function yourls_api_action_version() {
 }
 
 /**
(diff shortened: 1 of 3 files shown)


LeoColomb:
Thanks for your contribution!
That said, please see #1511. 👀

werpat:
Oh I didn't see that one. Thanks for pointing it out!

PR #2167, opened 2016-11-18
We need to run this over https

skynet:
We need to run this over https (see Google's recommendations for 2017)

Statistics page for a certain link throws a warning in the console:

```
Mixed Content: The page at 'xxxxxxxxxx' was loaded over HTTPS, but requested an insecure image 'http://www.google.com/s2/favicons?domain=xxxxxxxxxxx'. This content should also be served over HTTPS.
```

### DIFF
--- a/includes/functions-infos.php
+++ b/includes/functions-infos.php
@@ -241,7 +241,7 @@ function yourls_get_domain( $url, $include_scheme = false ) {
  *
  */
 function yourls_get_favicon_url( $url ) {
-	return yourls_match_current_protocol( 'http://www.google.com/s2/favicons?domain=' . yourls_get_domain( $url, false ) );
+	return yourls_match_current_protocol( 'https://www.google.com/s2/favicons?domain=' . yourls_get_domain( $url, false ) );
 }
 
 /**


ozh:
I don't get it. If you run on HTTPS, the icon is loaded from HTTPS. Example: https://yourls.org/dm+ and http://yourls.org/dm+

skynet:
Not really, link is hard-coded. Look at the code: https://github.com/YOURLS/YOURLS/pull/2167/files

skynet:
Oh I see now:

https://github.com/YOURLS/YOURLS/blob/bb4a764b65ee8d54b162dc4447e59d4853758a16/includes/functions.php#L1930

skynet:
If you don't patch it I'll have to add this to /user/config.php:

```php

// Find out if this is requested over SSL and set global variable
if ((isset ( $_SERVER ['HTTP_FRONT_END_HTTPS'] ) && $_SERVER ['HTTP_FRONT_END_HTTPS'] == 'On') || (isset ( $_SERVER ['HTTP_X_FORWARDED_PROTO'] ) && $_SERVER ['HTTP_X_FORWARDED_PROTO'] == 'https')) { // NETSCALER handles SSL
	$_SERVER ['HTTPS'] = 'on';
}

```

LeoColomb:
@skynet You should prefer to make a plugin to add your small latest patch with `HTTP_FRONT_END_HTTPS` handling. 
See in [`yourls_is_ssl` function](https://github.com/YOURLS/YOURLS/blob/bb4a764b65ee8d54b162dc4447e59d4853758a16/includes/functions.php#L1790) :
```php
return yourls_apply_filter( 'is_ssl', $is_ssl );
```
But you got it! 😺 

I close this PR since it's can be resolved with a simple plugin.

ozh:
Anyway, the PR was incorrect, since it would not change the outcome.
And anyway * 2, that function `yourls_match_current_protocol()` is dumb. It should be deprecated and we should just add external resources using `//link` notation and just not giving a f* about the protocol

dgw:
> we should just add external resources using //link notation and just not giving a f* about the protocol

:+1: Yes, that's the solution.

PR #2190, opened 2017-01-25
Set logout_link to fully empty

DatDraggy:
No more empty list element on the public index because $logout_link is now fully empty
https://puu.sh/tzCUh/28ce052988.png

### DIFF
--- a/includes/functions-html.php
+++ b/includes/functions-html.php
@@ -759,7 +759,7 @@ function yourls_html_menu() {
 	
 	// Now output menu
 	echo '<nav role="navigation"><ul id="admin_menu">'."\n";
-	if ( yourls_is_private() && !empty( $logout_link ) ) //Always displayed because $logout_link wasn't empty on public/domainroot index
+	if ( yourls_is_private() && !empty( $logout_link ) ) 
 		echo '<li id="admin_menu_logout_link">' . $logout_link .'</li>';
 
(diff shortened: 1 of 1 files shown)


ozh:
I don't get the purpose of this.

Replacing `$link = yourls_apply_filter( 'logout_link', '' )` with `$link = ''` gives basically the same (ie `''`) except that in the first case there is a filter so a plugin can modify the value, whereas when the value is hardcoded you can't.

And the other line changed is just about adding a comment

So, unless I grossly overlook something, I don't see the point of this

DatDraggy:
Look at the screenshot. When you have the version with the filter in, it doesnt make it fully empty so there will be an empty list element on the public yourls root index. That you can even interact with if you hover your mouse over it.

Yea the comment is pretty useless

DatDraggy:
Okay yea I didnt really think about plugins there... I guess something needs to be changed in the apply filter function so that the filter can stay there

ozh:
The thing is, there has to be something else doing something in your code. I cannot replicate this with the default public interface sample provided, and anyway your patch doesn't change anything:

```
var_dump(yourls_apply_filter('logout_link', '' )); // original code. Output: string '' (length=0)
var_dump(''); // your code. Output: string '' (length=0)
```

You probably have something with a filter on `'logout_link'`.

PR #2211, opened 2017-03-13
Fix action links not appearing to screen readers (fixes #2209)

### LINKED ISSUE #2209: accessibility: issues with screen readers on the admin dashboard
### Technical details

* YOURLS version: 1.7.1
* PHP version: 7.1.2
* Any other useful information depending on context (server version, mysql version, browser version, OS version...) Windows 7, Firefox, Jaws for windows (tested).

### Reproducible Bug Summary 

1. Load a screen reader (I used jaws for windows, as that's my screen reader I use -- there is a demo available).
2. Browse to yoursite/admin, and log in if necessary.
3. Try to navigate the table. You may notice the screen reader jumping around the contents seemingly randomly.

This is a bug because... It makes it very difficult to view the entire table of links (and possibly other items).

lilmike:
I changed the way the links are hidden, so that screen readers will still see them, but sighted users, until the css that usually would run visibility:visible, runs. This is because screen readers will not hover or anything over the actions column.

See http://webaim.org/techniques/css/invisiblecontent for the idea.


ozh:
Does not work for me: action links are rendered as empty vertical divs (FF, Chrome, Win7)
image

Also, some CSS declarations are broken in your PR: `left-10000px;`, missing `:`

lilmike:
Hi,
I basically changed display:none to a format that still shows to screen readers, because they don't hover over anything. I will fix the missing:, but I'm not sure, unless I misunderstood what it was supposed to do to begin with, what could be wrong about the CSS causing a blank div. just to make sure, the intended behavior is that the links don't show up until you hover over them, right?
Michael.

Sent from my iPhone

On Apr 4, 2017, at 11:41 AM, ྅༻ Ǭɀħ ༄༆ཉ <notifications@github.com<mailto:notifications@github.com>> wrote:


Does not work for me: action links are rendered as empty vertical divs (FF, Chrome, Win7)
[image]<https://cloud.githubusercontent.com/assets/223647/24668127/103aacb4-1966-11e7-9742-e1ceef9292f7.png>

Also, some CSS declarations are broken in your PR: left-10000px;, missing :

ozh:
@lilmike yeah at least it's currently the expected behavior. Action links show when you hover anywhere on the row (not just the action links themselves)

lilmike:
I just fixed the missing : character. Does that fix it? If not, I'm honestly stumped -- I'm not a css expert by any means, and I just followed the instructions given in the link above (which is generally considered a good source on accessibility on the web). If it doesn't work now, I will attempt to get more information.
-Michael.

ozh:
Nope -- same exact results as previously

lilmike:
I've just gotten around to looking at this further, as I was updating my yourls instances. I've changed it from using absolute positioning on the left to clipping it inside a 1x1 rectangle. This allows screen readers to see the links just fine, but only allows sighted users to see it whenever the visibility:visible would have run in the css (by doing positioning static and clip auto).
Let me know if this fixes the problem of not being able to see the links for sighted users.

lilmike:
Ok, also added aria markup to the action links when they're disabled or enabled to let screen readers know.

lilmike:
@ozh, @LeoColomb, any feedback? Can the links be seen by sighted users when hovering?
-Michael.

lilmike:
Huh, apparently my email didn't post to the thread. Anyway, what would be the best positioning here to "undo" absolute positioning. According to the w3schools position attribute reference page, static is the default.

-Michael.

ozh:
Definitely not working for me, now the action icons are all visible yet each on top of each other (win10, FF, Chrome)
image

I'm closing this PR is fairly old so I'm not sure anyone is still alive, but I'd really be glad with some newer feedback or PR from @lilmike or @audreymaniez maybe?

PR #2213, opened 2017-03-20
FIX the redirection after login in the admin panel

### LINKED ISSUE #2206: yourls_sanitize_url_safe() breaks redirect after admin login
### Technical details

* YOURLS version: 1.7.2
* PHP version: 5.6

### Reproducible Bug Summary 

1. Install YOURLS
2. Visit /admin Login page
3. Enter valid credentials
4. Browser is redirected to http://admin/? instead of going to http://YOURDOMAIN/admin

Expected behaviour: Show YOURLS admin page.

Cause: This change is responsible for the problem: https://github.com/YOURLS/YOURLS/commit/6748ce8a947e38e5b411ca6729967aa47b3fae9f#diff-de33dce122935846f969c7ff08091a8e

Replacing yourls_redirect( yourls_sanitize_url_safe($url) ); with yourls_redirect( $url ); fixes the problem.

Preovaleo:
Fix #2206 
By adding the Global variable `YOURLS_SITE` to the redirection

### DIFF
--- a/includes/functions-auth.php
+++ b/includes/functions-auth.php
@@ -80,7 +80,7 @@ function yourls_is_valid_user() {
 			
 			// Login form : redirect to requested URL to avoid re-submitting the login form on page reload
 			if( isset( $_REQUEST['username'] ) && isset( $_REQUEST['password'] ) && isset( $_SERVER['REQUEST_URI'] ) ) {
-				$url = $_SERVER['REQUEST_URI'];
+				$url = sprintf("%s%s", YOURLS_SITE, $_SERVER['REQUEST_URI']);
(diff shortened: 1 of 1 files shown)


ozh:
Ah, indeed, you're onto something.

Anyway this won't work when YOURLS is installed in a subdir. Also, this needs to comply to the current SSL situation (ie if URL requested is https, return https)

ozh:
I think this should work in all cases:
`yourls_match_current_protocol(yourls_sanitize_url(sprintf("%s%s", $_SERVER['SERVER_NAME'], $_SERVER['REQUEST_URI'])))`

(albeit not the best readable thingie)

PR #2234, opened 2017-04-20
Nitpicks

ozh:
Minor and useless code readability improvements

### DIFF
--- a/includes/functions-html.php
+++ b/includes/functions-html.php
@@ -193,7 +193,22 @@ function yourls_html_addnew( $url = '', $keyword = '' ) {
  * @return string Result
  */
 function yourls_html_tfooter( $params = array() ) {
-	extract( $params ); // extract $search_text, $page, $search_in ...
+    // Manually extract all parameters from the array. We prefer doing it this way, over using extract(),
+    // to make things clearer and more explicit about what var is used.
+    $search       = $params['search'];
+    $search_text  = $params['search_text'];
+    $search_in    = $params['search_in'];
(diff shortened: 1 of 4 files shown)


includes/functions-html.php
```diff
@@ -193,7 +193,22 @@ function yourls_html_addnew( $url = '', $keyword = '' ) {
...
+    $sort_by      = $params['sort_by'];
+    $sort_order   = $params['sort_order'];
+    $page         = $params['page'];
+    $perpage      = $params['perpage'];
+    $click_filter = $params['click_filter'];
+    $click_limit  = $params['click_limit'];
+    $total_pages  = $params['total_pages'];
+    $date_filter  = $params['date_filter'];
+    $date_first   = $params['date_first'];
+    $date_second  = $params['date_second'];
```
LeoColomb: May raise some errors on old and/or custom installations, since keys are hard-required now.

ozh: "keys are hard-required" = ? :)

LeoColomb: Ha ha, yes, some neologism mixing hard-coded and required! 😸 
Some calls of this function (in plugins?) may not include all this keys...
This not a big problem, but I just think it's good to keep in mind that it may break some user plugins.

ozh: Nah, I thought of this and I think we're safe, since this function call is [not filtered](https://github.com/YOURLS/YOURLS/blob/master/admin/index.php#L310-L325), so plugins should not be able to interfere here

PR #2252, opened 2017-05-18
Move 3rd party stuff in /vendor

ozh:
:feelsgood:


ozh:
Now to update the tests as well ...

ozh:
Man that's beautiful:
image

PR #2258, opened 2017-05-29
Add hook for plugins to add meta tags

dgw:
I also fixed the referrer meta tag being indented with spaces instead of a tab, since I was in the neighborhood. :)


ozh:
🦄

PR #2261, opened 2017-06-09
Support optional www. prefix

mle86:
Some people just insist on typing "www." into every URL.
YOURLS does not support that at the moment if `YOURLS_SITE` does not have the www\. prefix.

With this change, both http[s]://sho.rt/xyz _and_ http[s]://<b>www.</b>sho.rt/xyz will work,
no matter if `YOURLS_SITE` has a www\. prefix or not.

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -1911,9 +1911,9 @@ function yourls_get_request() {
 		return $request;
 	
 	// Ignore protocol & www. prefix
-	$root = str_replace( array( 'https://', 'http://', 'https://www.', 'http://www.' ), '', YOURLS_SITE );
+	$root = str_replace( array( 'https://www.', 'http://www.', 'https://', 'http://'  ), '', YOURLS_SITE );
 	// Case insensitive comparison of the YOURLS root to match both http://Sho.rt/blah and http://sho.rt/blah
-	$request = preg_replace( "!$root/!i", '', $_SERVER['HTTP_HOST'] . $_SERVER['REQUEST_URI'], 1 );
+	$request = preg_replace( "!(?:www\.)?$root/!i", '', $_SERVER['HTTP_HOST'] . $_SERVER['REQUEST_URI'], 1 );
 
 	// Unless request looks like a full URL (ie request is a simple keyword) strip query string
 	if( !preg_match( "@^[a-zA-Z]+://.+@", $request ) ) {


dgw:
This looks like something that should be handled by the webserver, IMO. Something as simple as this nginx example:

````nginx
server {
    listen 80;
    listen 443 ssl;
    # ssl cert directives here/included
    server_name www.sho.rt;
    return 301 $scheme://sho.rt$request_uri;
}
````

Having YOURLS itself special-case the `www` subdomain doesn't make sense to me (though a plugin might be able to do it, I don't think it belongs in core).

**Comment by mle86** [comment]
I see your point, although it looks like YOURLS is already _trying_ to special-case the `www.` prefix – see [functions.php:1914](https://github.com/YOURLS/YOURLS/blob/6642e951656a5ee9121145bc7858a79104de9fdd/includes/functions.php#L1914).

(That `str_replace` call however does not do what it is supposed to, as the search array is in the wrong order.)

**Comment by ozh** [comment]
What do you mean "wrong order"? It replaces any of `'https://'`, `'http://'`, `'https://www.'` or `'http://www.'` with an empty string, ie "Ignore protocol & www. prefix" as stated in the comment

**Comment by mle86** [comment]
No, it doesn't.
`str_replace` finds and replaces the search strings in the order they're given – it does not try to find the longest match first.
So for the input string `"http://www.sho.rt"`, the first match will be `'http://'` which will be removed.
After that, the rest of the search array is checked, but resulting string will be `"www.sho.rt"`, for which the search array contains no matches anymore.
See https://3v4l.org/l1ZYq

As it is right now, the third and fourth search strings will never be evaluated.

**Comment by ozh** [comment]
In-fucking-deed! Obviously. Duh. Nice catch.

But then why add `(?:www\.)?` in the regexp, since `www.` is always stripped out?

**Comment by mle86** [comment]
If you change the array order, then the `"www."` prefix gets removed from the `YOURLS_SITE` configuration constant – but only there.
It won't be removed from the user-input `HTTP_HOST` domain which the regex processes, and that means that the `preg_replace` result might be something like `"www./123"`.

That means that an input url with `www.` will never be considered valid again, no matter if the user includes the prefix in the configuration or not.
Two ways around that: either clean up the search array (`['https://', 'http://']`) to keep the current behavior (but without the misleading third and fourth search strings) in which the `YOURLS_SITE` domain has to match the `HTTP_HOST` exactly,
or modify the regex to ignore any `www.` prefix in the input.

**Comment by ozh** [comment]
Yeah, I really overlooked all this, thanks for the heads up.
Your fix works as expected for all scenarios, ie :

```php
$tests = array(
    //array('YOURLS_SITE', 'hostname', 'uri'),

    array('http://sho.rt', 'sho.rt', '/blah'),
    array('https://sho.rt', 'sho.rt', '/blah'),
    array('http://www.sho.rt', 'sho.rt', '/blah'),
    array('https://www.sho.rt', 'sho.rt', '/blah'),

    array('http://sho.rt', 'www.sho.rt', '/blah'),
    array('https://sho.rt', 'www.sho.rt', '/blah'),
    array('http://www.sho.rt', 'www.sho.rt', '/blah'),
    array('https://www.sho.rt', 'www.sho.rt', '/blah'),
);
```

I'll implement this and add some [unit tests](https://github.com/YOURLS/YOURLS-unit-tests) as well

@dgw yeah this can (and probably should) be solved at server config time, but I'm afraid Average Joe isn't really savvy here

ozh:
Aaaand I even manage to click "close" right after stating "this is cool". Don't github after a sleepless night :)

dotLuxury:
Friends don’t let friends ‘git with no sleep… ;-)


Best Regards,
A2

. L   U   X   U   R   Y

Travel.Luxury<https://travel.luxury/?utm_channel=email&utm_source=A2_corp_email> • Fashion.Luxury<https://fashion.luxury/?utm_channel=email&utm_source=A2_corp_email> • Discover.Luxury<https://discover.luxury/?utm_channel=email&utm_source=A2_corp_email>
Search.Luxury<https://search.luxury/?utm_channel=email&utm_source=A2_corp_email> • On.Luxury<https://on.luxury/?utm_channel=email&utm_source=A2_corp_email>

facebook: Dot Luxury<https://www.facebook.com/Dot-Luxury-1415136685408608/> • instagram: dotLuxury<http://instagram.com/dotLuxury>
twitter: @dotLuxury<http://twitter.com/dotLuxury>

dgw:
@ozh I was being idealistic, wasn't I? :)

And you might want to edit @dotLuxury's comment down to remove the GitHub unsubscribe links (with your tokens in, I believe), email headers/quote, and billboard-sized signature…

PR #2276, opened 2017-08-04
Fix logout in stats page

eli0tc:
I noticed that I couldn't logout when I was on the stats page.
I changed the relative path to a full path using yourls_admin_url() and now it works.

### DIFF
--- a/includes/functions-html.php
+++ b/includes/functions-html.php
@@ -743,7 +743,7 @@ function yourls_html_menu() {
 
 	// Build menu links
 	if( defined( 'YOURLS_USER' ) ) {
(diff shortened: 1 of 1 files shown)


LeoColomb:
This looks great! :sparkles:
Thanks for your contribution @eli0tc!

What is your motivation for this change?

eli0tc:
The link rendered looked like "https://domain.name/installation_path/uuuu+?action=logout'' which had no effect.

ozh:
I think this is specific to nginx, I can totally logout while on a stat page.

This causes a change of behaviour (as of now you stay on the stats page and are presented the login form ; your patch redirects you from the stats page to the admin area) but I don't think that's a big deal. I'm merging this, thanks for your PR.

It's just annoying that nginx seems to handle cookies in such a peculiar way (there's an ongoing issue with login on nginx). Out of curiosity, on a URL like `/bleh+?action=omg`, what does `$_REQUEST` look like on your setup? I don't get why [this](https://github.com/YOURLS/YOURLS/blob/master/includes/functions-auth.php#L16-L21) would fail...

eli0tc:
$_REQUEST is empty.
It seems that no get parameters are passed.

PR #2282, opened 2017-08-18
Remove ezSQL, use PDO

ozh:
Work in progress

# Todo:

* [x] remove ezSQL, use PDO functions with the Aura.SQL wrapper lib. Woot.
* [x] make all \YOURLS\Database\YDB properties `protected`
* [x] write ezSQL compatible functions (`$ydb->get_results()` and such), mark them deprecated
* [x] rewrite all SQL queries to use PDO
* [x] pay attention to all `@todo` and `@tempnote` left in the process
* [x] Write SQL upgrade docs in the wiki

This is not related to PDO, but for the lulz, option function logic was moved to its own class, to leave very simple wrapper functions instead.

In the end, next to `/includes/vendor`, we should have `/includes/YOURLS` with a few classes for most complex functions

## Regarding wrapper functions vs class functions

The idea is to :
- make wrapper functions as small & concise as possible
- keep the complex logic (validation, tests...) out of sight, in the class
- keep return filters inside wrappers
- move internal filters / actions in the class

Typical `yourls_` function eventually, in `/includes/functions*.php` :

```php
function yourls_do_stuff($var) {
    // Allow plugins to short-circuit options
    $pre = yourls_apply_filter( 'shunt_wrapper_for_stuff', false, $var );
    if ( false !== $pre )
        return $pre;

    global $ydb;
    $value = \YOURLS\Do\Stuff($ydb, $var, $bleh);

    return yourls_apply_filter('do_stuff', $value);
}
```


LeoColomb:
> PDO functions with the Aura.SQL wrapper lib

Why choosing Aura.Sql? 😉 

> all SQL queries to use PDO

Actually this is why I don't understand your changes: since we're talking about database wrappers refactoring, why not abstracting queries and objects? Then the code could be fully independent of the driver.

> move internal filters / actions in the class

You're moving to decorators and observers, right?

ozh:
@Leo : 
- Aura.SQL: well, why not. I looked into Doctrine and DBAL, I didn't wanted to add 800 KB of files for features I won't use, I even considered just using plain PDO, but found that Aura has a minimal set of interesting features to make things easier. It's merely PDO + just a few extra things.
- Abstraction : at the moment, I'm focusing on making something fully compatible with the existing code to make sure no 3rd party stuff (plugins, public shorteners, whatever tool using YOURLS) will break. So basically I want all the old `$ydb->get_row()` or any other ezSQL legacy function work (albeit throwing a "deprecated" notice). Then I'll rewrite all YOURLS queries, in particular using PDO binding : at this step yes abstracting things would be nice, although I didn't really think about it yet. If you have any idea or POC on how it could be intelligently done, I'm all ears :)
- Decorators and observers: I have no idea what you're talking about... Again, any code piece to illustrate what you mean is welcome.

ozh:
All unit tests pass, f* yeah.
300full

LeoColomb:
C'est pas facile de bosser avec toi en collab', j'aurai bien aimé relire avant la merge ! 😉

ozh:
@LeoColomb arf désolé je pensais pas que tu suivais :)

PR #2300, opened 2017-09-28
Adding label to every input on index page

audreymaniez:
None of the inputs were correctly labelled, it's a problem for screen reader user (e.g. a blind person), so I try to fix it in the less invasive way : 

- Add `label` tag and id/for relation to input in the new_url form : https://www.w3.org/TR/WCAG-TECHS/H44.html
- Add `aria-label` to every input in the sort form : https://www.w3.org/TR/WCAG20-TECHS/ARIA6.html

No CSS impact, and normally, no cross browser impact.

### DIFF
--- a/includes/functions-html.php
+++ b/includes/functions-html.php
@@ -181,8 +181,8 @@ function yourls_html_addnew( $url = '', $keyword = '' ) {
 	<div id="new_url">
 		<div>
 			<form id="new_url_form" action="" method="get">
-				<div><strong><?php yourls_e( 'Enter the URL' ); ?></strong>:<input type="text" id="add-url" name="url" value="<?php echo $url; ?>" class="text" size="80" placeholder="http://" />
-				<?php yourls_e( 'Optional '); ?> : <strong><?php yourls_e('Custom short URL'); ?></strong>:<input type="text" id="add-keyword" name="keyword" value="<?php echo $keyword; ?>" class="text" size="8" />
+				<div><label for="add-url"><strong><?php yourls_e( 'Enter the URL' ); ?></strong></label>:<input type="text" id="add-url" name="url" value="<?php echo $url; ?>" class="text" size="80" placeholder="http://" />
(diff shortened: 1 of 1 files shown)


ozh:
Did you use existing translatable strings, or did you introduce new ones?

audreymaniez:
Hey!

No, i introduce new string cause none of those available fit my needs. Is that a problem ? I try not to, but I didn't see any way to do without creating new ones....

But it seams the PR failed ^^

includes/functions-html.php
```diff
@@ -359,8 +365,8 @@ function yourls_html_tfooter( $params = array() ) {
  * @param boolean $display false (default) to return, true to echo
  * @return string HTML content of the select element
  */
-function yourls_html_select( $name, $options, $selected = '', $display = false ) {
-	$html = "<select name='$name' id='$name' size='1'>\n";
+function yourls_html_select( $label, $name, $options, $selected = '', $display = false ) {
```
LeoColomb: I would not add the label as the first argument. Also, it may be good to add default empty value for compatibility.

ozh:
Nah, the PR failed because currently the Unit Test suite is synced with another YOURLS branch. No worries, your changes won't introduce anything that will break the build.

This said, as @LeoColomb said, since you're introducing a new parameter in function calls, it should be added last, otherwise this would break any existing code.
eg : 
```php
// before:
yourls_html_something( $name, $options, $selected = '', $display = false ) { ... }

// after:
yourls_html_something( $name, $options, $selected = '', $display = false, $label = '' ) { ... }
```

ozh:
Ping @audreymaniez -- just a reminder if you'd like to update your PR so we can merge it

audreymaniez:
Hey, Thank you for the reminder. I think it's ok now ?

ozh:
Still not sure about this variable coming out of the blue ...

audreymaniez:
Ah... Not sure of its utility ? or not sure that adding a variable is the best way to do ?

Maybe there is a better way to do that, but adding a variable seemed to me the less invasive way. Not adding variable mean restructure a little bit more the code. Doing an other way will modify considerably the translation files (modify string instead of adding new ones, and i'm not sure it is better for your maintenance and future udpate ?).

ozh:
@audreymaniez : I don't see in the code where this var is defined ?

PR #2310, opened 2017-10-25
Skip title search for URLs to accelerate bulk import process

hamza-s63:
To accelerate bulk import process, we use the long url as the title which reduce the overall time significantly. Further tested on 12,000+ bulk import urls works perfectly in less than 2 minutes.


ozh:
You're suggesting to change the way YOURLS works just for the case when you are importing URLs. I don't think this is relevant.

What you want is a plugin to temporary (or permanently, depending on your needs) disable this feature.

PR #2314, opened 2017-11-03
Fix "find config"

ozh:
realpath() seems to have many caveats depending on the platform (Win/Linux/BSD) and various settings (namely, open_basedir restrictions)

Ping @boldcity and @KabataMacharia, if you could test this patch on your platform it would be very great. Currently works on Windows and Ubuntu

### DIFF
--- a/includes/YOURLS/Config/Config.php
+++ b/includes/YOURLS/Config/Config.php
@@ -23,7 +23,7 @@ class Config {
      * @param  mixed $config   Optional user defined config path
      */
     public function __construct($config = false) {
-        $this->set_root( $this->fix_win32_path(realpath(__DIR__ . '/../../../..')) );
+        $this->set_root($this->fix_win32_path( dirname(dirname(dirname(__DIR__))) ));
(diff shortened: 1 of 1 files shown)


LeoColomb:
@ozh It's already fixed with 4cb866a, isn't it?

ozh:
@LeoColomb see https://yourls.org/admin/

leachim6:
This fixed the issue for me running nginx  1.13.6-1~stretch and  php 7.0+49

andrewperry:
I just had to install this patch to get this repo working on Debian 8

ozh:
Yeah apparently `realpath()` isn't the most cross platform function...

ozh:
(for the record and a future quick refactor: it would be more logical if `YOURLS\Config\Config`'s constructor had two params, say `$root` and `$config`, with `$root` defining what we're guessing the line after with that `dirname()` avalanche. Would be more logical to pass that param to the class instead of practically hardcoding it within the class)

PR #2315, opened 2017-11-10
Add `docker`

kamilzabielski:
This commit provides support for `Docker`.
It provides `Dockerfile` and same of `docker-compose.yml`.
It also shows an example of cofiguration file that handles environment variables: `config-sample-docker.php`.

To run the whole `yourls` stack, you need to type:
* `docker-compose up`,
in root git-archive directory.

In case of issues that were mentioned by `codacy/pr`.
* `MAINTAINER` is deprecated [[1]]; `LABEL` should be used instead,
* `--no-install-recommends` is a kind of default for `debian` based images, but added this parameter.

[1]: https://docs.docker.com/engine/reference/builder/#maintainer-deprecated

Signed-off-by: Zabielski, Kamil <kamil.zabielski@outlook.com>


LeoColomb:
Hi @kamilzabielski!
Thanks for taking the time to submit a pull request, very interesting submission! 🎉 

That said, I would rather prefer docker configuration to be in another repository.
See very good examples:
* https://github.com/elastic/elasticsearch-docker
* https://github.com/docker-library/wordpress
* https://github.com/docker-library/drupal

Also, I found [many other Docker images for YOURLS](https://hub.docker.com/search/?q=yourls), could be really interesting to contribute to one of them (and maybe import the repo into the @YOURLS org?).

What is your thought? 😃

kamilzabielski:
I can totally agree on having separate repository, but, obviously, I was not able to do so. :-)

My proposal is to prepaire an official image and get rid of these forks.
The reason I did not decide to contribute to different one:

* They do not keep proper conventions and best practices.
* They either use some strange shell-scripts or does not keep _KISS_ practice.
* Most of them install senseless stuff, like supervisord or anything else,
* Some of them have senseless dependencies, like _New relic_, what should not happen.
* ...

ozh:
I'm all for having an "official" Docker image, that's a neat idea.

This said, I agree with @LeoColomb, this would be best in a separate repo, to be used by the people would use Docker, one idea being to avoid cluttering the main repo with something that most users don't care about.

Also, I think your approach is sane (not enabling random plugins, having odd dependencies, keeping things KISS and such) and I think this should be the goal of an official image, letting users fork it and tuning it to their liking.

I've created https://github.com/YOURLS/Docker for such an image : please feel free to submit a pull request there and let your idea begin this repo. We'll see how things evolve from there

PR #2318, opened 2017-11-16
add Dockerfile for easy deployment

tschuy:
This is the Dockerfile that I've been using for production; it's been working quite nicely and has the benefit of using the upstream php image to manage the installation of php and the various required php extensions.

Other alternative Dockerfiles I'd considered included [daftlabs](https://github.com/daftlabs/docker-yourls) and [tsgoff](https://github.com/tsgoff/docker-yourls), but both were unnecessarily complicated, and I figured it'd be good to have a canonical file included in the repository.

If y'all accept this, I'd happily add some documentation on setting up a local instance of YOURLS using Docker to the wiki.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@
 # Configuration
 user/*
 !user/config-sample.php
+!user/config-env.php
 .htaccess
 *.config
 *.conf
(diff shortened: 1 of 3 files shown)


ozh:
Please see #2315 and discussion.

PR #2323, opened 2017-11-26
Update placeholder to https://

flokX:
and motivate users to use the secure version of the url.


ozh:
This is the input box where people enter the URL to shorten.

dgw:
@ozh Yes, and I actually think it's a reasonable place to show `https://` as a subtle way to maybe get users to check if their link is available over HTTPS if it isn't already the secured version.

LeoColomb:
I second you @dgw. I think also it's worth adding it.

ozh:
The people have spoken!

PR #2325, opened 2017-11-27
First Probot config

LeoColomb:
This Pull Request aims to add a Probot config.
This should help maintainers to mitigate issues and comments management with a basic but automated workflow.

Application based on https://github.com/bkeepers/workflow
See also https://probot.github.io/ for reference

### DIFF
--- a/.github/MISSING_ISSUE_TEMPLATE_AUTOREPLY.md
+++ b/.github/MISSING_ISSUE_TEMPLATE_AUTOREPLY.md
@@ -1,0 +1,4 @@
+Hey @{{ sender.login }}, thanks for opening an issue!
+
+Unfortunately, it seems you are missing information from the [issue template](/.github/ISSUE_TEMPLATE.md).
+Please provide all nexxessary information as described in our [contribution guidelines](/.github/CONTRIBUTING.md).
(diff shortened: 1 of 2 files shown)


.github/MISSING_ISSUE_TEMPLATE_AUTOREPLY.md
```diff
@@ -0,0 +1,4 @@
+Hey @{{ sender.login }}, thanks for opening an issue!
+
+Unfortunately, it seems you are missing information from the [issue template](/.github/ISSUE_TEMPLATE.md).
+Please provide all nexxessary information as described in our [contribution guidelines](/.github/CONTRIBUTING.md).
```
ozh: "necessary"
Also, some issues are not issues (support requests, feature requests, etc, or very simplistic but accurate bug reports) so maybe reword the message with something like "make sure it has all the necessary information as described in ..."

.github/probot.js
```diff
@@ -0,0 +1,25 @@
...
+    .trim()
+    .replace(/\s+/gi, ' ')
+    .split(' ')
+    .length > 200
+  )
+  .comment(`
+Thanks for your comment, @{{ sender.login }}!
+
+Your reply is very long, and even if for me it takes less than a second to
+read it, summarizing it a bit will help old-fashioned humans. :smiley:
```
ozh: yes :)

PR #2331, opened 2017-12-04
Exception - Cannot find config.php - Wrong Directory Bug

PopVeKind:
#  Subdirectory Install

The /YOURLS/ directory was removed but the dirname() structure was not
changed. Throws exception - Cannot find config.php - in subdirectory
install.


LeoColomb:
Yes, that's my fault with https://github.com/YOURLS/YOURLS/commit/c8ad874cdf32997df8a793baffd403bffa04596e.
Good catch and thanks for the pull request!

PR #2336, opened 2017-12-10
Security: Enhance default .htaccess settings

guessi:
it is easy to detect if the site is running YOURLS via
```
- http[s]://<url>/LICENSE
- http[s]://<url>/.github/CONTRIBUTING.md
- http[s]://<url>/images/yourls-logo.png
- http[s]://<url>/css/...
- http[s]://<url>/js/...
```
and find out the running version via site traversing
```
- http[s]://<url>/readme.html
- http[s]://<url>/composer.json
- http[s]://<url>/composer.lock
- http[s]://<url>/sample*.txt
```
if there's any vulnerability exposured, it could be dangerous

better to hide those sensitive info by providing secure .htaccess

Signed-off-by: guessi <guessi@gmail.com>

### DIFF
--- a/includes/functions-install.php
+++ b/includes/functions-install.php
@@ -105,12 +105,47 @@ function yourls_create_htaccess() {
 		// Prepare content for a .htaccess file
 		$content = array(
 			'<IfModule mod_rewrite.c>',
-			'RewriteEngine On',
-			'RewriteBase '.$path.'/',
-			'RewriteCond %{REQUEST_FILENAME} !-f',
-			'RewriteCond %{REQUEST_FILENAME} !-d',
-			'RewriteRule ^.*$ '.$path.'/yourls-loader.php [L]',
+			'  RewriteEngine On',
+			'  RewriteBase '.$path.'/',
+			'  RewriteCond %{REQUEST_FILENAME} !-f',
+			'  RewriteCond %{REQUEST_FILENAME} !-d',
+			'  RewriteRule ^.*$ '.$path.'/yourls-loader.php [L]',
 			'</IfModule>',
+			'', // empty line
+			'# avoid file listing',
+			'<IfModule mod_autoindex.c>',
+			'  Options -Indexes',
+			'</IfModule>',
+			'', // empty line
+			'# avoid config content leak',
+			'<Files config*.php>',
+			'  Order Allow,Deny',
+			'  Deny from All',
+			'</Files>',
+			'', // empty line
+			'# avoid information leak',
+			'<Files sample*.txt>',
+			'  Order Allow,Deny',
+			'  Deny from All',
+			'</Files>',
+			'', // empty line
+			'# avoid running app version exposure',
+			'<Files composer*>',
+			'  Order Allow,Deny',
+			'  Deny from All',
+			'</Files>',
+			'', // empty line
+			'# avoid running app version exposure',
+			'<Files LICENSE>',
+			'  Order Allow,Deny',
+			'  Deny from All',
+			'</Files>',
+			'', // empty line
+			'# avoid information leak by file extension',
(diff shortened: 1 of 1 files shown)


ozh:
Also via //url/admin, //url/yourls-api.php, and so on.
Also, hackers don't check if site is running &lt;SOFTWARE> or what version of &lt;SOFTWARE> before trying an exploit. They just save a query and directly try whatever exploit.

Endless measures and counter measures and counter-counter measures. I don't think there's a point in doing this.

This said, other people than you might find this interesting, so I would suggest you make a plugin to "enhance" the generated htaccess.

guessi:
Yes, I agree that hacker would do that trick without knowing running app,

but I'm not agree with providing the "simple" default settings,
especially for not protecting "compose.json", I don't think it should be done by "plugin".

providing secure example is obviously easier way than "plugin" for most people.

ozh:
@LeoColomb toughts on adding a couple files (`composer.*` & `LICENSE`) to `.gitattributes`'s `export-ignore`? Arguably those won't be needed to people who download a zip.

dgw:
@ozh I think it's always a good idea to exclude files from zip/tarball if not needed for build/deployment, fwiw

LeoColomb:
To exclude the license, why not, even if distributing a package with its license is always recommended. 🆗 

Concerning `composer.json`, excluding this file, will disable any composer installation or composer usage as a distribution system. 👎 

Regarding the whole PR, many of these new rules are supposed to be defined in a server level, not at the application level. You can always [add this kind of rules](https://github.com/h5bp/server-configs-apache) on the app level by your own, but this is up to you.
Also, be careful @guessi, usage of `Order` & `Deny` rules is [deprecated](https://httpd.apache.org/docs/current/mod/mod_access_compat.html).
😺

guessi:
to be honest, it doesn't feel good to close PR without discussion 👎 

what I'm trying to do is initiate a new conversation about current default settings,
that's why I start this PR originally,

@ozh @LeoColomb I totally agree with providing package with LICENSE, but still, it shouldn't be exposed to world by default, especially when it is running under production mode,

@dgw I would say exclude only `sensitive` files/dirs only, and of course, with proper configure

@LeoColomb I think you might get my words wrong, I'm not trying to remove them when packaging, but not to expose to world.

and yes, I'm aware that `Order, Allow, Deny` is getting deprecated, and it is always better to configure at server level, but it did provided a basic `.htaccess` through `includes/functions-install.php`, why not make it better by default?

LeoColomb:
@guessi I can totally understand your feelings, and I'm sorry about that.
That said, a closed PR is not a closed discussion (we are discussing right now), and a closed pr can even be reopened. 
And @ozh motivated the close of the pull request, thus explained why this pull request does not comply with his vision of YOURLS development. 
I also added my review, so we got two arguments against:
* These changes do not improve security in general;
* It's not a common practice to include such rules by default, specifically at the application level, in the world of php-httpd apps.

As a side note, please accept my [advice](https://opensource.guide/how-to-contribute/#-your-contribution-doesnt-get-accepted) to not ever take a pr rejection personally. Even the best contributors can make a non-relevant change regarding the project direction! 👿 
You said you wanted to start a discussion about sensitive configuration, this is a really good idea, we are all open to this and I even encourage you to do this! 😘

guessi:
@LeoColomb thanks for update

- this PR simply hide sensitive data, to prevent being exposure to world, as I explained before
- although there's no big security issue right now (or being discovered), but one day it will... 🔨  
- every web app (npm, yarn, godeps, ...) will have the same concern for the package.{json|lock} being leak
- for the current design (<=1.7.2), it did really easy to discover that `Ohh... it's running YOURLS`
- for sure, this PR doesn't improve that much, or not a general practice, but it's start a conversation :-)

I totally agree that it should be done by server side (apache, nginx, etc.)

but there's an existed application logic (functions-install.php) for detecting if `.htaccess` exist or not, and it did provide a simple runnable .htaccess,

I still insist that provided .htaccess file should be stronger, even it could lower down system performance

PR #2345, opened 2018-01-04
add filter to bypass yourls_html_addnew()

joshp23:
This adds a simple filter allowing a plugin to bypass the yourls_html_addnew() function. This is useful if an alternate form with special request parameters needs to be used.


joshp23:
I worked out how to accomplish my goal with javascript, however, I think that this is still a better way to go about things, so I'll leave this open. Cheers.

PR #2357, opened 2018-01-14
Share tweaks

### LINKED ISSUE #2352: Twitter 140 character limit
Twitter has updated their character limit to 280 characters now. Would be nice if the share dialog reflected this.

dgw:
This is two separate small tweaks to the `share.js` code, one of which fixes #2352 (the other is just because I was in the file already and saw it).

One or the other can be cherry-picked, or I can just merge both. If both, I'll probably squash it to one commit before merging because the changes are both minor and proximate.

@JayHoltslander this is the PR I promised, a couple days later than planned but not forgotten!

### DIFF
--- a/js/share.js
+++ b/js/share.js
@@ -9,8 +9,8 @@ $(document).ready(function(){
 function update_share() {
 	var text = encodeURIComponent( $('#tweet_body').val() );
 	var url = encodeURIComponent( $('#copylink').val() );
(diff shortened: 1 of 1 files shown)


dgw:
Guess I won't squash then 😹

ozh:
I have stopped a long time ago paying attention and putting efforts into a cleaner commit log. I think that's a sad waste of time. 2 commits for 2 tiny stuff? No problem :)

PR #2362, opened 2018-01-26
Fixed broken save when URL is left blank

dagee3:
Fixed issue where you could leave a url blank when editing and it would allow it as a value.  Now will verify if it is blank before allowing it to be processed.


ozh:
I think the idea is sane, indeed. Not so sure about the implementation : `yourls_add_new_url()` is already an awful mess that's pretty difficult to read, adding a `elseif(){ } else { }` is really adding reading complexity. At this point I think we should rearrange things a bit.

dagee3:
I can look at this later in the week when I have some spare time.  I think I can help reorganize it to make it more readable.

ozh:
Back to this. I don't see why we need to introduce new functions? Will fix with a simpler approach.

PR #2376, opened 2018-03-19
Update functions-auth.php

ymage:
3. Using defined constant is sufficient


rossigee:
I've just come up with the same fix for the same problem. LGTM!

ozh:
I don't think this is going to work when YOURLS is installed in a subdirectory

rossigee:
It should work just fine if YOURLS_SITE is defined with the subdirectory part included, which it should be if properly configured.

dgw:
Superseded by #2438.

Issue #2383, opened 2018-04-10
mixed content on stats page  


vineettalwar:
- [x] Check you are using the [**LATEST** release](https://github.com/YOURLS/YOURLS/releases/latest) or the development branch.
- [x] Make sure you have [**SEARCHED** existing issues](https://github.com/YOURLS/YOURLS/issues?utf8=%E2%9C%93&q=is%3Aissue) first. 
- [x] Read the [**GUIDELINES**](https://github.com/YOURLS/YOURLS/blob/master/.github/CONTRIBUTING.md) linked in the yellow notice box above.
- [x] **COMPLETE** this issue template.

### Technical details regarding my environment

* YOURLS version: 1.7.2
* Plugins enabled: none 
* PHP version: 5.5.9
* Any other useful information depending on context (server OS & version, MySQL version, browser version, …)
OS: Ubuntu 14.04.5 LTS
nginx/1.4.6

### Bug description 
- Mixed Content: The page at 'https://domain.name/url+' was loaded over HTTPS, but requested an insecure image 'http://www.google.com/s2/favicons?domain=domain.name'. This content should also be served over HTTPS.
- A similar issue and similar server setup like https://github.com/YOURLS/YOURLS/pull/2167

**What is the current behavior?**
console throws mixed content on stats page 

This is a bug because…
- hardcoded url in functions-info.php:243 "http://www.google..."

**What is the expected or desired behavior?**
- Without hard coded protocol and using //link 

**Debugging summary**
-  I did debug the is_ssl as well, 
- var_dump($_SERVER['HTTPS']) is returns null, despite ssl certificate is set . Thus, is_ssl returns false.

dgw:
`$_SERVER['HTTPS']` not being set properly is a server configuration issue. YOURLS matches protocols when it detects that it's running over HTTPS, so you'd probably be well served to find the configuration problem and fix it. That will prevent other mixed-content errors and such from both YOURLS and any other apps that check whether SSL is enabled.

PopVeKind:
PS you did not say what you are trying to use https for? Example, some people use https for **only the admin** area but not the short URLs.  Other people want to use https for **both admin and short URLs**.  Most people use http for everything!

LeoColomb:
> Most people use http for everything!

Not really, using HTTP without TLS in production is now highly discouraged.

PopVeKind:
@LeoColomb - Hey Leo, this might make a very good point of discussion.  Let's start with a few quotes...

[...]

PR #2384, opened 2018-04-10
Mixed content issue on stats page

vineettalwar:
Removal of hard coded http  and using //link notion to fetch favicon
This solves #2383

### DIFF
--- a/includes/functions-infos.php
+++ b/includes/functions-infos.php
@@ -241,7 +241,7 @@ function yourls_get_domain( $url, $include_scheme = false ) {
  *
  */
 function yourls_get_favicon_url( $url ) {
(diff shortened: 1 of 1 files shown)


dgw:
There's no reason to use a protocol-independent URI. Just always request this resource over HTTPS.

But the real question is, why doesn't `yourls_match_current_protocol()` do its job on your setup? I'll post more in #2383 about that, because it shouldn't be happening anyway.

LeoColomb:
Thanks for your contribution @vineettalwar, but I second @dgw, this seems more related to `yourls_match_current_protocol()`.

Closing for now, may be re-opened if needed after #2383 resolution.

PR #2415, opened 2018-08-01
Fixed link to contributors graph.

MatthewC:
I simply fixed the URL to the contributors graph, before it was giving me a 404 error, because it was taking me here: https://github.com/YOURLS/YOURLS/blob/master/graphs/contributors


MatthewC:
No problem :)

PR #2438, opened 2018-09-16
Update functions-auth.php

ymage:
3. Using defined constant is sufficient
4. Address concern from #2376

### DIFF
--- a/includes/functions-auth.php
+++ b/includes/functions-auth.php
@@ -80,7 +80,8 @@ function yourls_is_valid_user() {
 			
 			// Login form : redirect to requested URL to avoid re-submitting the login form on page reload
 			if( isset( $_REQUEST['username'] ) && isset( $_REQUEST['password'] ) && isset( $_SERVER['REQUEST_URI'] ) ) {
-				$url = yourls_match_current_protocol(yourls_sanitize_url(sprintf("%s%s", parse_url(YOURLS_SITE, PHP_URL_HOST), $_SERVER['REQUEST_URI'])));
+                $yourls_url = parse_url( YOURLS_SITE );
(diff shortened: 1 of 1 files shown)


rossigee:
This doesn't work for me. Why should we only be interested in the hostname part of the here? Aren't the protocol, port, and path parts equally important here? 

For example, my YOURLS_SITE is 'https://mydomain.com/'. Because this patch continues to ignore the protocol part (https), we end up with a 'Location: mydomain.com/blahblah' header, which causes the browser to send the user to a site which doesn't exist. Fail.

Similarly, if I were to host it on a subdomain, such as 'https://mydomain.com/yourls/', that'd ignore the path part of the YOURLS_SITE settings too, and also redirect the user to the wrong place. Fail.

Why not just use YOURLS_SITE, as configured?

ymage:
I believe subdirectory is handled by $_SERVER['REQUEST_URI']. 
wdyt ?

rossigee:
That looks like it would work, but I still fail to understand why you feel the need to parse and reconstruct the URL? What is the problem with just using the YOURLS_SITE constant as it is?

ymage:
Well, I think @Ozh is probably right about subdirectory install. And this would be handled twice : in YOURLS_SITE and in REQUEST_URI

rossigee:
That's true. I see the point, now.  :+1: from me.

ozh:
This one will fail in the case YOURLS_SITE has a port specified (eg `https://sho.rt:81/`).

In the end we need to have the original URL requested that needed auth, whatever its format and scheme (eg `http://sho.rt/admin/tools.php` or `https://mysite.com/yourls/admin/index.php` or `http://johan@lanserver:1337/admin`)

ozh:
Good to patch while being iso, better to improve the whole stuff :)
As you want regarding using this PR or opening a new one (if I understood your question correctly)

PR #2439, opened 2018-09-17
Add top, bottom, and end hooks to the login form

### LINKED ISSUE #2436: Login form has no hooks
I spent a little time looking around to see how one might go about implementing a plugin to provide multi-factor authentication (as requested in #2434) or OpenID login (I have a WordPress plugin that consumes OpenID), for example. In the process, I could not find any way for plugins to add fields on the login form.

Perhaps YOURLS doesn't need quite as many hooks [as WordPress has](https://codex.wordpress.org/Customizing_the_Login_Form#Login_Hooks), but I do think at least one injection point (below username & password, but above the submit button) would be useful. A plugin implementing two-factor auth would add a field there asking for the TOTP code (for example); a plugin implementing OpenID as a consumer would add, say, "or" and a field for the user's OpenID URL.

Per usual, before I actually get into any code, I'd like to flesh out exactly what hooks are appropriate. We can copy WordPress's hook name `login_form` for the one at the bottom of the login form. It's also trivial to add a `login_head` hook in case plugins need to add tags in the header for CSS/JS that's needed only on the login page. Putting a `login_footer` hook might not be as useful…

Left to my own devices, I'd implement the first two (`login_head` and `login_form`) as a start.

dgw:
Plugins can use these to add extra fields, for OpenID login, accepting two-factor authentication codes, etc.

Fixes #2436. I'm not super attached to these hook names; they just made sense as a set. Happy to change them if someone else thinks of clearer names, or if they don't fit well enough with the existing hooks.

### DIFF
--- a/includes/functions-html.php
+++ b/includes/functions-html.php
@@ -727,6 +727,7 @@ function yourls_login_screen( $error_msg = '' ) {
 				if( !empty( $error_msg ) ) {
 					echo '<p class="error">'.$error_msg.'</p>';
 				}
+				yourls_do_action( 'login_form_top' );
 			?>
 			<p>
 				<label for="username"><?php yourls_e( 'Username' ); ?></label><br />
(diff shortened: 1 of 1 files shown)


dgw:
Test failure seems to be from HTTPbin flakiness, as each build job on Travis shows slightly different failures all in the Requests batch. What can you do, I guess… (Besides running a local HTTP server during the test phase instead of calling out to an external service, that is.)

ozh:
Yeah, httpbin flakiness can be a pain, but usually it just works fine enough :)

PR #2442, opened 2018-09-24
Replace unsafe dynamic $_SERVER['SERVER_NAME'] by safer YOURLS_SITE constant in login redirect

ymage:
3. Using defined constant is safer
4. Handles scheme, port and in URL username & password

### DIFF
--- a/includes/functions-auth.php
+++ b/includes/functions-auth.php
@@ -80,7 +80,12 @@ function yourls_is_valid_user() {
 			
 			// Login form : redirect to requested URL to avoid re-submitting the login form on page reload
 			if( isset( $_REQUEST['username'] ) && isset( $_REQUEST['password'] ) && isset( $_SERVER['REQUEST_URI'] ) ) {
-				$url = yourls_match_current_protocol(yourls_sanitize_url(sprintf("%s%s", $_SERVER['SERVER_NAME'], $_SERVER['REQUEST_URI'])));
+				$yourls_url = parse_url( YOURLS_SITE );
+				unset($yourls_url['path']);
+				unset($yourls_url['query']);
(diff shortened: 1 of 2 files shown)


PopVeKind:
Although I agree that `$_SERVER['SERVER_NAME']` can cause problems in an **improperly configured server**, the `YOURLS_SITE` constant currently also causes problems for newbies.

A better replacement might be `$_SERVER['HTTP_HOST']`as this depends on the request from the client.

I would vote to reject this PR as written.  After my extended health absence, I have started to write the PR and Plugin to solve **ALL** the Protocol Problem (http|s) and I believe it will solve this issue also.

Likewise, the cpnstants `YOURLS_SITE` and `YOURLS_ADMIN_SSL` will be removed from the core in this coming PR to prepare for a fast and easy **Full-Fledged Graphical Installer** and a true **YOURLS Multisite** (one set of files and more than one domain with different keywords per domain).  I expect the **_Protocol Problem PR_** and **_HTTPS Plugin_** will be complete and tested by December. (I currently have an _Alpha Prototype_ under test and am refining the code.)

ymage:
Any news ????

PopVeKind:
I was side tracked by a stroke last year.
SERVER_NAME is only unsafe due to bad coding use, not the element itself.

The real way forward is to remove all core constants except for the
database entries. Starting with the YOURLS_SITE constant.

This should be split into two database entries (one to build short URLs,
and one to build admin urls). These should be modified by a plugin for
people who do not need the default settings. I had this 95% coded when I
had my stroke.

ymage:
Hi,
Sorry to hear that.
Hope you recovered well.

Issue #2430, opened 2018-09-10
yourls_get_favicon_url is abused by yourls_match_current_protocol


tobozo:
yourls_get_favicon_url() seems to take for granted that PHP will always share the same protocol as the frontend server.

Problem: generated favicon protocol doesn't match admin console protocol when Yourls is behind a reverse proxy.

![image](https://user-images.githubusercontent.com/1893754/45296744-5297a780-b503-11e8-847b-19c7b1bd7c0b.png)

Steps to reproduce: 
1) Yourls is behind nginx reverse proxy: e.g. Apache/PHP (HTTP) <=> Nginx (HTTPS) <=> Client
2) Visit `https://my.yours.installation/xxx+` to see the stats of the `xxx` short URL

When these conditions are met, the favicon address to the long URL domain is generated using HTTP `http://www.google.com/s2/favicons/.....` instead of HTTPS, in some situations the favicon won't show up on the stats page.

Also a few warnings about mixed content can be observed in the console.

![image](https://user-images.githubusercontent.com/1893754/45296953-026d1500-b504-11e8-8644-51ac4c3db691.png)

Would it make more sense to use a protol relative address instead: `//www.google.com/s2/favicons/.....` ?

Using the following code made the warnings go away and the favicons show up: 

```
function yourls_get_favicon_url( $url ) {
        return yourls_match_current_protocol( '//www.google.com/s2/favicons?domain=' . yourls_get_domain( $url, false ) );
}
```
(I only removed the hardcoded "http:" from the URL)

ozh:
Anyway the whole 'yourls_match_current_protocol' stuff is deprecated now. We should probably just use scheme-less URL everywhere (eg [//url.com](https://www.paulirish.com/2010/the-protocol-relative-url/) and let the client decide if this is `http` or `https`)

tobozo:
> We should probably just use scheme-less URL everywhere (eg //url.com and let the client decide if this is http or https)

That stands with google favicon where both http and https are available, but there are still some http-only servers out there, and there are plugins that implement other protocols so it could be an issue to implement that "everywhere".

vineettalwar:
I had the similar issue, checkout #2383 . Solution for Leo colomb worked for me in that case.

PR #2445, opened 2018-10-10
Use HTTPS to get favicon from google service

MarcProe:
Using http will break https installs, as browsers will report a "mixed sources" issue and mark the whole page as unsecure.
Google provides the favicon service on http as well as on https, so it can be easily switched over to the https version.

### DIFF
--- a/includes/functions-infos.php
+++ b/includes/functions-infos.php
@@ -241,7 +241,7 @@ function yourls_get_domain( $url, $include_scheme = false ) {
  *
  */
 function yourls_get_favicon_url( $url ) {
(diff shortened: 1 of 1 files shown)


ozh:
Thanks for contributing! This said! this change won't have any effect: `yourls_match_current_protocol()` replaces `http` with `https` if you're on https.

Anyway, this function should be deprecated I think and replaced with just including scheme-less URLs in general, there's a discussion about it on another ticket or PR.

And more specifically for this favicon, what @dgw said :)

MarcProe:
Sorry, but something is wrong here. The patch _has_ an effect.

---
Without the patch:
image
image
---
With Patch:
image
image

MarcProe:
However, of course you are correct that this exact issue is addressed in #2430.

PR #2466, opened 2018-11-25
Fix grammar and typo in readme.html

leduyquang753:
The current readme.html contains at least two typing mistakes:
- "355 digits number" should be "355-digit number".
- "sofware" should be "software".

Pulling this request fixes them.


dgw:
Good catches, both. Thank you!

PR #2486, opened 2019-01-08
Replace deprectaed create_function with anonymous function

J0WI:


ozh:
Indeed, but shouldn't this be `function() use ( $message ) { echo "$message";}` instead of single quotes?

J0WI:
Fixed

ozh:
I don't have time right now to test this (and obviously unit test results show this isnt covered) but I'm a bit concerned you're submitting one or another without seemingly testing either :) I'll test when I have some time.

J0WI:
The goal of this is to officially support PHP 7.2 (and maybe also 7.3) as your Docker image already does.

includes/functions-html.php
```diff
@@ -839,7 +839,7 @@ function yourls_html_menu() {
 function yourls_add_notice( $message, $style = 'notice' ) {
 	// Escape single quotes in $message to avoid breaking the anonymous function
 	$message = yourls_notice_box( strtr( $message, array( "'" => "\'" ) ), $style );
-	yourls_add_action( 'admin_notices', create_function( '', "echo '$message';" ) );
+	yourls_add_action( 'admin_notices', function() use ( $message ) { echo "$message";} );
```
LeoColomb: Actually no quote is needed at all. 
Just add space before the bracket.

```suggestion
	yourls_add_action( 'admin_notices', function() use ( $message ) { echo (string) $message; } );
```

includes/functions-html.php
```diff
@@ -828,7 +828,7 @@ function yourls_html_menu() {
 	/*
 	To display a notice:
 	$message = "<div>OMG, dude, I mean!</div>" );
-	yourls_add_action( 'admin_notices', create_function( '', "echo '$message';" ) );
+	yourls_add_action( 'admin_notices', function() use ( $message ) { echo "$message";} );
```
LeoColomb:
```suggestion
	yourls_add_action( 'admin_notices', function() use ( $message ) { echo (string) $message; } );
```

J0WI: Thanks for taking the time to review this. I've applied the suggested changes.

ozh:
:+1:

PR #2518, opened 2019-03-21
feature: append hash with current timestamp to nonce generation

franzwilding:
Nonce should be unique on every request. Before the change, the nonce would only change when YOURLS_NONCE_LIFE expires. This allows attacker to sniff the communication and discover token for specific action. This will allow to perform further Cross-Site Request Forgery attacks.

With my change, the token will be re-generated every time the user reloads the page


ozh:
I don't think this works as intended : while it does generate a unique nonce every time (and I think that's an interesting idea), previously generated nonces are still verified as true. In your scenario (attacker sniffing a nonce) this would not change anything.

ozh:
The more I think about this, the more I'm convinced the approach is actually flawed. Unless you store each nonce in the DB every time it's generated (which would add SQL overhead), I don't think you can have a time dependent string that can be invalidated by generating another string.

franzwilding:
The nonces will be invalid, once YOURLS_NONCE_LIFE expires. Until this, a new nonce will be generated on each request, making it at leat unique. 

I think the best solution would be to make an ajax request first to get a nonce (https://en.wikipedia.org/wiki/Cryptographic_nonce) which would allow us to dramatically reduce YOURLS_NONCE_LIFE to just 10 seconds or something. However, this might be an overkill here

ozh:
Issuing SQL queries on each nonce generation is way too far stretched IMHO for core inclusion. I think the best from there would be to make a plugin for those who want extra security and don't care about extra SQL queries

ozh:
@stefan-kamsker just to be sure I read this correctly : does your commit revert *all* @franzwilding changes and add filter on top of original core code ?

stefan-kamsker:
@ozh 
exactly, we solved this with an plugin, but need to alter nonce generaton, all changes from @franzwilding  are reverted

PR #2519, opened 2019-03-24
Fix yourls_api_stats() with filter==rand

### LINKED ISSUE #2516: Action `stats` with filter `rand` gives and SQL Exception
* [x] I've read the [guidelines for Contributing to YOURLS Projects](https://github.com/YOURLS/.github/blob/master/CONTRIBUTING.md)
* [x] This request isn't a duplicate of an existing issue
* [x] I've read the [docs](https://yourls.org/) and followed them (if applicable)
* [x] This is not a personal support request that should be posted on the [YOURLS Discourse](https://discourse.yourls.org/) community

## Description

SQL Exception caused by using the action `stats` with filter `rand`.

## Steps to reproduce

Using httpie

```bash
http get https://some-server-name.com/yourls-api.php signature==44aa822a05dfe6d33f36d3b07297b39c timestamp==1552654863 action==stats limit==3 filter==rand format==json
```

I always get this response

```
HTTP/1.1 500 Internal Server Error
Connection: keep-alive
Content-Type: text/html; charset=UTF-8
Date: Fri, 15 Mar 2019 13:06:19 GMT
Server: nginx/1.14.0 (Ubuntu)
Transfer-Encoding: chunked
```

**Expected behavior:** Get stats of 3 random urls.

**Actual behavior:**

```
PHP message: PHP Fatal error:  Uncaught PDOException: SQLSTATE[42S22]: Column not found: 1054 Unknown column 'RAND()' in 'order clause' in /path/to/YOURLS-1.7.3/includes/vendor/aura/sql/src/ExtendedPdo.php:748
Stack trace:
#0 /path/to/YOURLS-1.7.3/includes/vendor/aura/sql/src/ExtendedPdo.php(748): PDOStatement->execute()
#1 /path/to/YOURLS-1.7.3/includes/vendor/aura/sql/src/ExtendedPdo.php(523): Aura\Sql\ExtendedPdo->perform('SELECT * FROM `...', Array)
#2 /path/to/YOURLS-1.7.3/includes/functions.php(599): Aura\Sql\ExtendedPdo->fetchObjects('SELECT * FROM `...')
#3 /path/to/YOURLS-1.7.3/includes/functions-api.php(168): yourls_get_stats('rand', 3, 0)
#4 /path/to/YOURLS-1.7.3/includes/functions-api.php(37): yourls_api_stats('rand', '3', '')
#5 /path/to/YOURLS-1.7.3/includes/functions-plugins.php(150): yourls_api_action_stats(false)
#6 /path/to/YOURLS-1.7.3/yourls-api.php(35): yourls_apply_filter('api_action_stat...', false)
#7 {main}
```

**Reproduces how often:** How often you make this API call

## Versions

- OS: Ubuntu 18.04.2 LTS
- YOURLS 1.7.3
- MySQL 8.0.13
- PHP 7.2.15
- PHP-FPM 7.2.15
- NGINX 1.14.0

### PHP Packages

- php-common
- php-pear 1.10.5
- php7.2-bcmath
- php7.2-cli
- php7.2-common
- php7.2-curl
- php7.2-fpm
- php7.2-gd
- php7.2-json
- php7.2-mbstring
- php7.2-mysql
- php7.2-opcache
- php7.2-readline
- php7.2-xml
- php7.2-zip

### NGINX

```
server {
  listen 443;
  # SSL Certificates configurations
  # ...
  server_name some-server-name.com;
  charset utf-8;

  root /path/to/YOURLS-1.7.3;
  index index.php;
  location / {
    try_files $uri $uri/ /yourls-loader.php$is_args$args;
  }
  error_page 403 = @main_site;
  location @main_site {
    return 301 https://some-other-site;
  }
  location ~ \.php$ {
    proxy_set_header X-Real-IP $remote_addr;
    include fastcgi.conf;
    fastcgi_index index.php;
    fastcgi_pass 127.0.0.1:9002;
  }
}
```

### PHP-FPM Pool

```
[yourls]
user = www-data
group = www-data
listen = 127.0.0.1:9002
listen.allowed_clients = 127.0.0.1
# PM Dynamic configuration, 2 start servers with a maximum of 5
# ...
chdir = /path/to/YOURLS-1.7.3
catch_workers_output = yes
```

ozh:
Fixes #2516

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -571,11 +571,11 @@ function yourls_get_stats( $filter = 'top', $limit = 10, $start = 0 ) {
 
 	switch( $filter ) {
 		case 'bottom':
-			$sort_by    = 'clicks';
+			$sort_by    = '`clicks`';
 			$sort_order = 'asc';
 			break;
 		case 'last':
-			$sort_by    = 'timestamp';
+			$sort_by    = '`timestamp`';
 			$sort_order = 'desc';
 			break;
 		case 'rand':
@@ -585,7 +585,7 @@ function yourls_get_stats( $filter = 'top', $limit = 10, $start = 0 ) {
 			break;
 		case 'top':
 		default:
-			$sort_by    = 'clicks';
+			$sort_by    = '`clicks`';
 			$sort_order = 'desc';
 			break;
 	}
@@ -596,7 +596,7 @@ function yourls_get_stats( $filter = 'top', $limit = 10, $start = 0 ) {
 	if ( $limit > 0 ) {
 
 		$table_url = YOURLS_DB_TABLE_URL;
-		$results = $ydb->fetchObjects( "SELECT * FROM `$table_url` WHERE 1=1 ORDER BY `$sort_by` $sort_order LIMIT $start, $limit;" );
+		$results = $ydb->fetchObjects( "SELECT * FROM `$table_url` WHERE 1=1 ORDER BY $sort_by $sort_order LIMIT $start, $limit;" );
 
 		$return = array();
 		$i = 1;
@@ -1063,7 +1063,7 @@ function yourls_upgrade_is_needed() {
 	// Check if YOURLS_VERSION exist && match value stored in YOURLS_DB_TABLE_OPTIONS, update DB if required
 	if( $currentver < YOURLS_VERSION )
 		yourls_update_option( 'version', YOURLS_VERSION );
-    
+
 	return false;
 }
 


includes/functions.php
```diff
@@ -596,7 +596,7 @@ function yourls_get_stats( $filter = 'top', $limit = 10, $start = 0 ) {
 	if ( $limit > 0 ) {
 
 		$table_url = YOURLS_DB_TABLE_URL;
-		$results = $ydb->fetchObjects( "SELECT * FROM `$table_url` WHERE 1=1 ORDER BY `$sort_by` $sort_order LIMIT $start, $limit;" );
+		$results = $ydb->fetchObjects( "SELECT * FROM `$table_url` WHERE 1=1 ORDER BY $sort_by $sort_order LIMIT $start, $limit;" );
```
LeoColomb: What is the idea behind `WHERE 1=1`?

dgw: It would make sense if this query also had a `$where_filter` (or whatever) value interpolated into it that might or might not contain additional conditions. Having it anyway is harmless (essentially zero effect on query execution time), but doesn't appear to serve any purpose in this particular case.

ozh: It has always been there ([see in 2009...](https://github.com/YOURLS/YOURLS/blob/1.2/includes/functions.php#L357)).

I guess it's either because at this time I couldn't dare writing an SQL clause without a WHERE, or because at this time there were supposed to be potential troubles omitting this WHERE, either with MySQL or other DB engines.

I think it's also for readability to clearly specify this query DOESNT have a WHERE clause

PR #2542, opened 2019-08-02
Fix Type Juggling Bypass in Auth Functions

Wocanilo:
The auth functions are vulnerable to Type Jugging attacks. Non strict comparisions can be used by an attacker to bypass auth and gain access to the admin page and API.

If PHP decides that both operands looks like numbers, it will convert both an perform numeric comparision. Ex: "0e348324" == "0e99" is true

This commit fixes 4 vulnerabilities:
- If an user password is saved as plaintext and looks like a number, for example, "0e23", an attacker can bypass login by submitting an input that looks like a number, "0e99".
- If an user cookie looks like a number, for example "0e210903587487700652447287594086", an attacker can bypass login by sending a cookie that also looks like a number, "0e32".
- If an user signature looks like a number, for example, "0e344332", an attacker can bypass API auth by sending a signature that also looks like a number, "0e84"

At last, the time limited token functionality can be abused to gain unauthorised access to the API without any requisite. Again, a type juggling attack is posible, but this time the attacker controls the result of the MD5, so thousands of requests can be made till the resultand hash takes the form of a number. We can bypass the YOURLS_NONCE_LIFE by introducing decimals in the timestamp, making the attack viable.

References:
[https://www.owasp.org/images/6/6b/PHPMagicTricks-TypeJuggling.pdf](https://www.owasp.org/images/6/6b/PHPMagicTricks-TypeJuggling.pdf)

### DIFF
--- a/includes/functions-auth.php
+++ b/includes/functions-auth.php
@@ -133,7 +133,7 @@ function yourls_check_password_hash( $user, $submitted_password ) {
 		return( $yourls_user_passwords[ $user ] == 'md5:'.$salt.':'.md5( $salt . $submitted_password ) );
 	} else {
 		// Password stored in clear text
-		return( $yourls_user_passwords[ $user ] == $submitted_password );
+		return( $yourls_user_passwords[ $user ] === $submitted_password );
 	}
 }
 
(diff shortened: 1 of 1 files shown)


skaag:
I have a test yourls installation that some bots were able to find and are once in a while injecting some malicious shortened URL into... I applied this patch, but it's still happening. Any ideas?

ozh:
@skaag most likely something else involved. URL of your test install ?

skaag:
> @skaag most likely something else involved. URL of your test install ?

This is it (It has just 3 shortened links on it as a test, works fine so far):
https://gns.to/

skaag:
@ozh :
Steps I took today: I changed the cookie, and the password for the only user in that system (and that user is not `admin`).

ozh:
Soooooo.... "bots" found and are "injecting" URL ... on your *publicly available* interface? Hmmkey.

PR #2545, opened 2019-08-07
Fixed redirect failure after login

hangxingliu:
This bug appears on the website isn't hosted on port 80/443.
Use `HTTP_HOST` to replace the `SERVER_NAME`.

For example, The `$_SERVER['SERVER_NAME']` maybe `127.0.0.1` when the website is hosted on `127.0.0.1:8008`. But `$_SERVER['HTTP_HOST']` will be `127.0.0.1:8008` in this situation.

### DIFF
--- a/includes/functions-auth.php
+++ b/includes/functions-auth.php
@@ -9,7 +9,7 @@ function yourls_is_valid_user() {
 	if ( null !== $pre ) {
 		return $pre;
 	}
-	
+
 	// $unfiltered_valid : are credentials valid? Boolean value. It's "unfiltered" to allow plugins to eventually filter it.
 	$unfiltered_valid = false;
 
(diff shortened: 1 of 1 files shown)


dgw:
Don't revert #2542.

This looks like it will _always_ add `:port` to the URL if `$_SERVER['HTTP_HOST']` is not set and the code has to fall back on `$_SERVER['SERVER_NAME']`. We really don't want or need it for vanilla `http://` on port 80 and `https://` on 443. It's more likely to confuse users than not.

hangxingliu:
> Don't revert #2542, please.

Sorry, I forgot to pull the upstream code before modify.

And I modify the code around generating redirect URL. It would not add `:80` and `:443` anymore.

ozh:
Back (after so long, sorry) at this : I'm not sure why we're not simply using YOURLS_SITE instead of this complex $_SERVER['SERVER_NAME'] juggling.

If anyone here still alive : https://github.com/YOURLS/YOURLS/issues/2603#issuecomment-589594461

ozh:
Trying to make things even simpler. I'm closing this PR and looking forward to fixing things with PR #2625, please report there

PR #2326, opened 2017-11-28
Import unit-tests inside the main repo

LeoColomb:
As discussed in #2301, it would be easier (and more common) to manage unit-tests directly in the main YOURLS repository.

The git log from YOURS-unit-tests repo is kept.

PR #2552, opened 2019-09-21
Fix test suite

ozh:
Sync with https://github.com/YOURLS/YOURLS-unit-tests/pull/10

Why do we have 2 repos, one for the software and one for the tests, again? :smile:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,12 @@
+sudo: false
 language: php
 
 env:
   global:
   - DB=mysql
+
+services:
+  - mysql
 
 matrix:
   include:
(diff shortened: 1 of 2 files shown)


ozh:
@LeoColomb so, as usual I'm poking here and there with poor attempts at fixing composer and things but it doesn't work. Any idea? :)

LeoColomb:
> Why do we have 2 repos

See https://github.com/YOURLS/YOURLS/pull/2326

> Any idea?

Path mismatches and old phpunit checks. 😉

ozh:
:+1: thanks !

PR #2554, opened 2019-09-25
Rewrite of social sharing

muellerlukas:
Hi there,

I had a problem with sharing on twitter (wrong/old url).
Sadly I saw no possibility to fix this with a plugin.
While coding I noticed a few other things

- tweet-limit in PHP was 140 (in PHP), now 280 everywhere
- enable/disable individual services (With fallback)
- instead only url, also text is posted to facebook (thats why I renamed "tweet_text" to "share_text"
- rewrote js: New url is only generated/written on click, not on every text-change (template in data-attribute). Url in href as non-js-fallback.

With the new js it is also more easy for plugin-devs to add other services. Just add a link with the href, the template and add the onclick.
Minimal example
```html
<a href="https://example.com/?t=<?php echo $text; ?>" 
data-shareurl="https://example.com/?t=#share#" 
onclick="share(this);return false;">TEST</a>
```


ozh:
Hello there, thanks for proposing that enhancement!

Looks good to me, except for the new global var added to user config : if anything, we're more prone to *remove* options that add ones. Those global vars are bad practice actually (difficult regarding to unit tests for instance)

It would be better to had filters and actions where needed to custom behavior (for the few who need it) can be set via plugin, instead of adding vars for everyone. For instance here, simply have the array of sharing services in an array hooked to a filter

muellerlukas:
Hi,

I later saw you pushed the PHP minimum-requirement to 5.6.
As of 5.6 supports arrays as constants. Maybe that would be a simple solution for defining new config-options?

muellerlukas:
Hi,

now I understand whar you mean.
I will remove the 2 default sharings (twitter, facebook) and build plugins for every single service. So every user can activate that sharings he want to use as a plugin. (Or none)

muellerlukas:
Example plugins (FB, Twitter): https://github.com/muellerlukas/yourls-socialshare

ozh:
I think you went further than I thought, sorry if I wasn't clear enough. I think out of the box things should be just as now : twitter and facebook sharing enabled. This will work for 99% of people. Plus, it's best practice not to hide things that were there for years, that will lead to dozens of issues and mails "why is the twitter thingy gone??"

But then there are some people who will want to add a network, and a few folks who will want to hide one or two. For those, having a plugin to add a sharing option (eg [Tumblr](https://github.com/YOURLS/YOURLS/wiki/Plugin-=-Share-Tumblr), by [mail](https://github.com/peterberbec/yourls-mailto), etc...) or hide one will be the best.

So the update to current sharing feature should enhance/fix things but not break what already works for people (ie people who already have a plugin, or people who just want things as default)

muellerlukas:
Ah, sorry. My fault. I misunderstood this.

Made some commits that includes fb/twitter again, removes the social-config and renames the field back to "tweet_body".
Other changes like the 280-limit and the new js usable for plugins remains.

I also added in the repo linked above 2 plugins to remove fb, twitter and some for adding new services

ozh:
Back on this, sorry for the delay.

Doesn't work for me: when I click on a Share icon, I just get empty links for Twitter and FB, eg `https://twitter.com/intent/tweet?text=` with no parameter.

This said, the PHP needs indeed a couple tweaks.

PR #2557, opened 2019-10-02
Fix composer warning about uppercase in require block

hnnweb:
Fix composer warning about uppercase in require block

```
$ composer install
Deprecation warning: require.rmccue/Requests is invalid, it should not contain uppercase characters. Please use rmccue/requests instead. Make sure you fix this as Composer 2.0 will error.
Loading composer repositories with package information
Installing dependencies (including require-dev) from lock file
Nothing to install or update
Generating autoload files
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,7 +20,7 @@
         "ext-pdo": "*",
         "ozh/bookmarkletgen": "1.2",
         "ozh/phpass": "1.2.0",
-        "rmccue/Requests" : "1.7",
+        "rmccue/requests" : "1.7",
         "pomo/pomo" : "1.4.1",
         "geoip2/geoip2" : "2.9.0",
         "aura/sql": "~2.",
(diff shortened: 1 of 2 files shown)


ozh:
Sure. I'll let @LeoColomb say if there is potential impact to this, I'm not even sure what we really need this composer file for (should the composer lock file be updated as well? other composer include files?)

hnnweb:
Updated the lock file. Ran `composer update nothing`

hnnweb:
It actually wrote 
```
	modified:   includes/vendor/autoload.php
	modified:   includes/vendor/composer/ClassLoader.php
	modified:   includes/vendor/composer/LICENSE
	modified:   includes/vendor/composer/autoload_real.php
	modified:   includes/vendor/composer/installed.json
```

this aswell. Want them too?

PR #2558, opened 2019-10-02
Split test in 2

ozh:
This test randomly fails on Travis. Splitting it in 2 to see which parts is inconsistent (I guess the one with mt_rand, but why?)


ozh:
Ha. Actually, both tests randomly fail. Heh.

PR #2574, opened 2019-11-05
Trim overlong client information (HTTP referrer and user-agent)

LorenzCK:
When following links from very long URLs (for instance, from mailing lists or newsletters), overlong client information fails to INSERT into the database because of the table's field lengths.
These two very simple changes trim overlong strings (maximum length as indicated in `includes\functions-install.php`).

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -888,8 +888,8 @@ function yourls_log_redirect( $keyword ) {
     $binds = array(
         'now' => date( 'Y-m-d H:i:s' ),
         'keyword'  => yourls_sanitize_string($keyword),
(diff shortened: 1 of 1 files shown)


ozh:
I can't believe this hasn't been fixed decades ago. Thanks !

(This said, user agents shouldn't be at fault, unless a plugin changes default : https://github.com/YOURLS/YOURLS/blob/master/includes/functions.php#L701)

LorenzCK:
> (This said, user agents shouldn't be at fault, unless a plugin changes default : https://github.com/YOURLS/YOURLS/blob/master/includes/functions.php#L701)

Didn't notice that! Anyway, better safe than sorry I guess. 😊 Thanks for the merge!

PR #2581, opened 2019-11-21
Allow sha1 for time limited signature token

Art4:
I noticed [here](https://github.com/YOURLS/YOURLS/wiki/PasswordlessAPI#usage-of-a-time-limited-signature-token) that md5 is used for time limited signature token.

This PR would allow sha1 hashs for time limited signature token but would be backward compatible. This is done by counting the string length of the signature. `md5()` creates a 32 char long hash, `sha1()` creates a 40 char long hash.

What do you think?


ozh:
This would need also doc update. I'm not sure this is really worth the hassle considering that the md5 is salted and time based ?

Art4:
I can do the doc updates as well if this will be get merged. I would change the docs to sha1 and mention the md5 as a BC feature. 

> I'm not sure this is really worth the hassle considering that the md5 is salted and time based ?

Since md5 is really broken it is just wrong to just use it without a secure alternative. Personally I would also prefer sha265 over sha1 because sha1 have several weaknesses too.

ozh:
So, let's do something better then.
In addition of allowing `yourls-api.php?timestamp=<time>&signature=<md5 hash>6&action=...`, I'll implement something to allow `yourls-api.php?timestamp=<time>&signature=<any hash>&hash=<hash function>&action=...`

Art4:
Good idea. With an implemantation with `hash_algos()` this might be ready for any future improvements in hash functions independent of the PHP version:

```php
if (in_array($_REQUEST['hash_algo'], hash_algos())) {
    $hash_algo = (string) $_REQUEST['hash_algo'];
} else {
    $hash_algo = "md5";
}

$result = (hash($hash_algo, $data) === $_REQUEST['signature']);
```

ozh:
that's the way I've implemented it

PR #2584, opened 2019-11-24
Changing the logic of obtaining data from the database

vit73:
This change was needed due to the long processing of data through includes /vendor/aura/sql/src/ExtendedPdo.php at the processing stage in Rebuilder

### DIFF
--- a/includes/Database/YDB.php
+++ b/includes/Database/YDB.php
@@ -481,5 +481,38 @@ class YDB extends ExtendedPdo {
         yourls_debug_log('LEGACY SQL: '.$query);
         return $this->fetchAffected($query);
     }
+
+    public function fetchValue($statement, array $values = array())
+    {
+        $pdo = parent::getPdo();
+        $pdo->setAttribute(PDO::MYSQL_ATTR_USE_BUFFERED_QUERY, false);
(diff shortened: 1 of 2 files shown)


ozh:
I'll need more than a one line explanation on this please

vit73:
During the execution of certain methods (for example, removing more than one million links from the database), a similar error occurs.

Fatal error: Allowed memory size of 134217728 bytes exhausted (tried to allocate 4096
bytes) in /var/www/html/includes/vendor/aura/sql/src/ExtendedPdo.php on line 748

During the analysis, a functional was discovered for processing data, which will later be connected via bindValue. But in our case, this processing is redundant and we had to redefine a number of functions so that the functionality that is too resource-intensive was not executed. Maybe there are some reasons why using the add-on over the add-on over PDO is connected? And in particular the need to use Rebuilder in /vendor/aura/sql/src/ExtendedPdo.php

vit73:
@ozh Is there any information on transferring changes to the main project?

vit73:
@ozh Is there any information on my previous comment and the initial question on updating the project?

ozh:
If I understand things correctly, this is specific to your use case. I'm closing the PR.

Issue #2307, opened 2017-10-12
utf8mb4 charset in MySQL causes error: Specified key was too long; max key length is 767 bytes


teyeheimans:
### Technical details regarding my environment

* YOURLS version: 1.7.3
  * Plugins enabled: none
* PHP version: PHP Version 7.0.22-0ubuntu0.16.04.1
* Using Aurora (Mysql Variant from Amazon) version 1.14 with default charset utf8mb4_general_ci

### Reproducible bug summary 

The installation fails and I receive a status code 500.  So there is no "friendly" error page. This is because there is no try/catch statement around the query. 

I found out that the query for the URL table fails because the primary key is too long: Error Code: 1071. Specified key was too long; max key length is 767 bytes. This is related because I use the default charset utf8mb4_general_ci

The create statement which is executed by yourls: 
```sql
CREATE TABLE IF NOT EXISTS `url` (
    `keyword` VARCHAR(200)BINARY NOT NULL,
    `url` TEXT BINARY NOT NULL,
    `title` TEXT CHARACTER SET UTF8,
    `timestamp` TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
    `ip` VARCHAR(41) NOT NULL,
    `clicks` INT(10) UNSIGNED NOT NULL,
    PRIMARY KEY (`keyword`),
    KEY `timestamp` (`timestamp`),
    KEY `ip` (`ip`)
);
```

For example, this also fails:
```sql
CREATE TABLE IF NOT EXISTS `url` (
    `keyword` VARCHAR(200) CHARACTER SET utf8mb4 COLLATE utf8mb4_general_ci  NOT NULL ,
    `url` TEXT BINARY NOT NULL,
    `title` TEXT CHARACTER SET UTF8,
    `timestamp` TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
    `ip` VARCHAR(41) NOT NULL,
    `clicks` INT(10) UNSIGNED NOT NULL,
    PRIMARY KEY (`keyword`),
    KEY `timestamp` (`timestamp`),
    KEY `ip` (`ip`)
) ENGINE=INNODB;
```

However, this works:
```sql
CREATE TABLE IF NOT EXISTS `url` (
    `keyword` VARCHAR(200) CHARACTER SET utf8 COLLATE utf8_general_ci  NOT NULL ,
    `url` TEXT BINARY NOT NULL,
[...]

Issue #1901, opened 2015-05-12
Allow multiple URL for the same reference


Nhoya:
Hey there, when i try to short an URL that already exist it gives me the previous shortned URL, there is a way shorten "already-shortned" links?

eg google.com -> http:// sh.rt/1a && http://sh.rt/344bfsha32

chtaube:
Sure. Just take a look into the configuration file `user/config.php` within your installation. Look for this:

```
/** Allow multiple short URLs for a same long URL
 ** Set to true to have only one pair of shortURL/longURL (default YOURLS behavior)
 ** Set to false to allow multiple short URLs pointing to the same long URL (bit.ly behavior) */
define( 'YOURLS_UNIQUE_URLS', true );
```

Read the comment block. Then edit the file:

Replace
```
define( 'YOURLS_UNIQUE_URLS', true );
```
with
```
define( 'YOURLS_UNIQUE_URLS', false );
```
Easy, isn't it? ;-)

Nhoya:
yes, thanks man!

PR #2593, opened 2020-01-13
Issue 2307 database creation fixes

Seegras:
Two fixes for https://github.com/YOURLS/YOURLS/issues/2307
- patch by teyeheimans to report db-creation errors
- patch by me to lower key size to 191, to accommodate mysql 5.6 databases using utf8mb4

### DIFF
--- a/includes/functions-install.php
+++ b/includes/functions-install.php
@@ -249,8 +249,19 @@ function yourls_create_sql_tables() {
 
 	// Create tables
 	foreach ( $create_tables as $table_name => $table_query ) {
-		$ydb->perform( $table_query );
-		$create_success = $ydb->fetchAffected( "SHOW TABLES LIKE '$table_name'" );
+	    try {
+            $ydb->perform($table_query);
+        }
+        catch ( Exception $e ) {
+            $error_msg[] = yourls_s(
+                "Error creating table '%s' because of error message: '%s'. Create statement: '%s'",
+                $table_name,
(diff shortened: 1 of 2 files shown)


ozh:
Thanks!
Can you elaborate on the 191 chars bit, to "accomodate utf8mb4 on mysql 5.6" ? What is it exactly?

ozh:
OK, read a bit about utf8mb4 and this (weird?) 191 char limit. Thanks a lot.

The PR needs a slight reworking: in order to trigger the DB upgrade, you need to bump the [DB version](https://github.com/YOURLS/YOURLS/blob/master/includes/version.php#L12) (arbitrary number, `483` will be fine), to [trigger a redirection](https://github.com/YOURLS/YOURLS/blob/master/includes/Config/Init.php#L92-L95) to the upgrade page. A newer function in `functions-upgrade.php` must reflect the newer DB version: don't change function yourls_upgrade_482(), create a newer one that will be run (`yourls_upgrade_483()` for instance)

Tell me if that's clear or if I can help

Seegras:
Oh, IC. But I'm sorry, I'm no longer interested. Now that it worked, it turned out you can only have one redirect per URL, which is a showstopper for me, so I looked at a different software.

dgw:
@Seegras You can, though. It's just a config change away. See #1901.

ozh:
Some of this PR is included in #2690

Issue #2605, opened 2020-02-14
nonce URLs are relative, not absolute - this affects proxying with a non-root path


pcolmer:
* [X] I've read the [guidelines for Contributing to YOURLS Projects](https://github.com/YOURLS/.github/blob/master/CONTRIBUTING.md)
* [X] This request isn't a duplicate of an existing issue
* [X] I've read the [docs](https://yourls.org/) and followed them (if applicable)
* [X] This is not a personal support request that should be posted on the [YOURLS Discourse](https://discourse.yourls.org/) community

**Describe the bug**
The URLs created by yourls_nonce_url are relative. They do not include the YOURLS_SITE value at the start. Normally, this would not be a problem but, if you are trying to proxy YOURLS with a non-root path, it is close to impossible to define the proxy rules correctly.

Changing the code to use absolute URLs greatly simplifies the proxy rules and does not affect normal usage.

**To Reproduce**
Configure Apache so that it proxies to docker-yourls with a non-root path, e.g.

```
ProxyPass /yourls/ http://localhost:8000/
ProxyPassReverse /yourls/ http://localhost:8000/
```

Now load admin/plugins.php and inspect the page source. The URLs for the actions are relative and so look like `/admin/plugins.php`. These will not work because all of the YOURLS URLs need to be proxied under /yourls/ in this example configuration.

**Expected behavior**
If YOURLS returned absolute URLs, `ProxyPass` and `ProxyPassReverse` would be to successfully rewrite the action URLs.

**Actual behavior**
The URLs are relative so the Apache path cannot get prepended. This breaks the interface.

**Versions**
1.7.6 in the Docker container.

ozh:
Closing to keep discussion going in the PR

PR #2606, opened 2020-02-14
Fix site URL reference

pcolmer:
This changes `yourls_add_query_arg()` so that it always returns absolute URLs. Since `yourls_nonce_url` calls `yourls_add_query_arg`, that seems to be the cleanest point to add the test for whether or not we have a relative URL and, if we do, prefix it with the YOURLS site URL.

This fixes https://github.com/YOURLS/YOURLS/issues/2605


pcolmer:
@LeoColomb 

> Just wondering, why the entire includes/functions-auth.php file is deleted?

It shouldn't have been. That was a mistake at my end in moving files from one directory to another instead of copying them.

I've restored the file.

ozh:
I'm totally not competent at having an opinion on this issue, since I've never used (reverse)proxies as described in 2605

I'll need to run all tests I can think about in various cases before having an idea on this. Won't be able to do so before at least 2 weeks though (dev box is away, only have access to github atm)

ozh:
OK, re-read this and the original issues you filed about nonces. The problem isn't about nonces or yourls_add_query_arg(), but is because we're passing relative URLs where we should pass absolute URLs. Closing here, but I will fix right now.

Issue #2607, opened 2020-02-17
Use of HTTP_HOST prevents path proxying against Docker container


pcolmer:
* [X] I've read the [guidelines for Contributing to YOURLS Projects](https://github.com/YOURLS/.github/blob/master/CONTRIBUTING.md)
* [X] This request isn't a duplicate of an existing issue
* [X] I've read the [docs](https://yourls.org/) and followed them (if applicable)
* [X] This is not a personal support request that should be posted on the [YOURLS Discourse](https://discourse.yourls.org/) community

**Describe the bug**
`yours_get_request` uses `HTTP_HOST` to divine the full URL of the shortlink request but, if yourls is running in a Docker container **and** it is being proxied from a non-root path, that logic is faulty.

**To Reproduce**
1. Set up a yourls Docker container with YOURLS_SITE set to the "external" FQDN URL **including** the "/app" path (see step 2).
2. Configure Apache to proxy "/app" to the underlying YOURLS instance.
3. Try to retrieve a short link.

**Expected behavior**
The underlying full link associated with the short link is returned.

**Actual behavior**
YOURLS fails to return the link because the use of the Apache variables causes the path to be left in the request field, thereby failing to match the shortlink.

**Versions**
1.7.6 in a Docker container.

ozh:
Closing to keep conversation in the referenced PR

PR #2604, opened 2020-02-14
Fix site URL reference

pcolmer:
This fixes https://github.com/YOURLS/YOURLS/issues/2603

PR #2608, opened 2020-02-17
Fix shortlink retrieval

pcolmer:
This fixes https://github.com/YOURLS/YOURLS/issues/2607.

It takes a similar approach to that taken in https://github.com/YOURLS/YOURLS/issues/2604 and in the `allow-aliases` plugin, resulting in no longer needing that plugin if yourls is being run in a Docker container with a correctly configured Apache front-end.

### DIFF
--- a/includes/functions-auth.php
+++ b/includes/functions-auth.php
@@ -80,7 +80,7 @@ function yourls_is_valid_user() {
 			
 			// Login form : redirect to requested URL to avoid re-submitting the login form on page reload
 			if( isset( $_REQUEST['username'] ) && isset( $_REQUEST['password'] ) && isset( $_SERVER['REQUEST_URI'] ) ) {
-				$url = yourls_match_current_protocol(yourls_sanitize_url(sprintf("%s%s", $_SERVER['SERVER_NAME'], $_SERVER['REQUEST_URI'])));
+				$url = yourls_match_current_protocol(yourls_sanitize_url(sprintf("%s%s", YOURLS_SITE, $_SERVER['REQUEST_URI'])));
 				yourls_redirect( yourls_sanitize_url_safe($url) );
 			}
 		}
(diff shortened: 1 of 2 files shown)


ozh:
This won't work when YOURLS is installed in a subdir

Example:
- YOURLS at http://sho.rt/yourls/
- requested page : http://shor.rt/yourls/abcd
- glueing YOURLS_SITE and REQUEST_URI will end up with http://sho.rt/yourls/yourls/abcd

pcolmer:
> This won't work when YOURLS is installed in a subdir

Confirmed. I'll try and fix a solution that works in all cases.

pcolmer:
When looking at how this runs inside a Docker container, calls to `yourls_get_request` result in values like this:

```
HTTP_HOST = localhost:8001
REQUEST_URI = /BoardRoomAV
```

I use the external value (FQDN) for `YOURLS_SITE` and that doesn't match `HTTP_HOST`, so the request gets munged into `localhost01/BoardRoomAV`. That then doesn't work so the redirects fail.

I'm struggling to get PHP logging working on a direct-installed instance of YOURLS. I'll update this when I've got that working.

ozh:
Could you test #2625 ?

pcolmer:
> Could you test #2625 ?

It doesn't work in a container for shortlink retrieval. It works for admin authentication and redirection afterwards but `yourls_get_request` needs some work. I think that function could perhaps be simplified and just look at the leaf part of the request URI to see if that is in the database.

But I can't be sure of my logic until I can get PHP logging to work :(

ozh:
Yeah, `yourls_get_request()` needs a makeover. I'll make something in 2625 since it's all connected to the same problem (aka "don't use $_SERVER stuff when you can do otherwise").

Closing here, to be continued in #2625

ozh:
@pcolmer I've simplified a lot yourls_get_request(), I think things should work on your side with my latest commit to PR 2625

Issue #2615, opened 2020-03-10
Allow webcal URLs to add


theknigge:
Can you please add protocol support for webcal/webcals URLs?
Like: webcals://link-to-my-calendar.com
Currently it appears "Missing or malformed URL"

That would be greate. Thank you in advance!

LeoColomb:
Thanks for opening this issue @theknigge.
You can use the related filter with a tiny plugin to add any protocol.

https://github.com/YOURLS/YOURLS/blob/6228a2874e532002286e99ecb364efb4eaa573fb/includes/functions-kses.php#L62

ozh:
Specifically, see https://github.com/YOURLS/custom-protocol

PR #2616, opened 2020-03-10
Whitelist webcal: + webcals protocol

theknigge:
Like in feature request #2615 described, webcal+webcals support would be greate.


ozh:
This would be plugin territory, see related issue

PR #2626, opened 2020-03-24
Fix test inconsistency in group `utils`

### LINKED ISSUE #2621: Tests inconsistency in group `utils`
Tests in `tests/utilities/next_decimal.php` randomly fail. This is weird. Must investigate.

ozh:
Simplify logic of tests in NextDecimal_Tests to only test what is supposed to be tested.

Should fix #2621


ozh:
Huh. I think I'm mixing two branches. Oh well... :)

PR #2627, opened 2020-03-24
Fix test inconsistency

### LINKED ISSUE #2621: Tests inconsistency in group `utils`

ozh:
Simplify logic of tests in NextDecimal_Tests to only test what is supposed to be tested.

(Replaces #2626 because I'm a git noob)

Should fix #2621

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -102,6 +102,8 @@ function yourls_get_IP() {
 /**
  * Get next id a new link will have if no custom keyword provided
  *
+ * @since 1.0
+ * @return int            id of next link
  */
 function yourls_get_next_decimal() {
 	return yourls_apply_filter( 'get_next_decimal', (int)yourls_get_option( 'next_id' ) );
@@ -110,6 +112,15 @@ function yourls_get_next_decimal() {
 /**
  * Update id for next link with no custom keyword
  *
+ * Note: this function relies upon yourls_update_option(), which will return either true or false
+ * depending if there has been an actual MySQL query updating the DB.
+ * In other words, this function may return false yet this would not mean it has functionnaly failed
+ * In other words I'm not sure we really need this function to return something :face_with_eyes_looking_up:
+ * See issue 2621 for more on this.
+ *
+ * @since 1.0
+ * @param integer $int     id for next link
+ * @return bool            true or false depending on if there has been an actual MySQL query. See note above.
  */
 function yourls_update_next_decimal( $int = '' ) {
 	$int = ( $int == '' ) ? yourls_get_next_decimal() + 1 : (int)$int ;
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -1125,25 +1125,10 @@ function yourls_get_option( $option_name, $default = false ) {
 		return $pre;
 
 	global $ydb;
-    $value  = $ydb->yourls_options->get($option_name, $default);
+    $option = new \YOURLS\Database\Options($ydb);
+    $value  = $option->get($option_name, $default);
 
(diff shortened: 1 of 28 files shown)


ozh:
Well at least now it's consistent ....

ozh:
30 000 tests on my dev box, cannot reproduce. FML.

ozh:
So, still haven't found the specific issue with these two tests. I've improved things here and there in the process of searching I think, so in the end if we drop this PR make sure we still cherry pick useful commits.

includes/functions.php
```diff
@@ -1158,9 +1171,7 @@ function yourls_get_all_options() {
  */
 function yourls_update_option( $option_name, $newvalue ) {
 	global $ydb;
-
-    $option = new \YOURLS\Database\Options($ydb);
-    $update = $option->update($option_name, $newvalue);
+    $update = $ydb->yourls_options->update($option_name, $newvalue);
```
ozh: @LeoColomb opinion on this? ie using the same `yourls_options` instead of making a new `Options($ydb)` each time? Useful? Useless? Meaningless optimization?

ozh:
So. I'm out of ideas. Still no clue why this happens sometimes, and not being able to reproduce on my machine despite the thousands of tests doesn't help.

I think I'm just going to remove these tests and live with it :-)

includes/functions.php
LeoColomb: Just weird! 😆
No, more seriously, it won't change anything.

Let's recap.

```php
new \YOURLS\Database\Options($ydb);
```
basically return `$ydb` wrapped up with some functions/methods:

https://github.com/YOURLS/YOURLS/blob/575ded5925352547ea498b93d888f19d3ba1197d/includes/Database/Options.php#L32-L34

In other words,

```php
$ydb->yourls_options = new \YOURLS\Database\Options($ydb);
```
is more or less equivalent to
```php
$ydb->yourls_options = $ydb; # <--- ouch
$ydb->yourls_options->update = function (...) {...}
```
which is not really useful.
And you still need to use the `global` (which is not really a best practice! 😅).

In terms of functionality, it's equivalent.
In terms of readability, opinion can differ I think.
In terms of brain visualization, it doesn't help at all.

---

Without your addition, a new object is created and destroyed at each test, but `$ydb` keeps the updates.
Then in terms of memory, it's almost equivalent.
And in terms of CPU, yaaaah, we might win some micro seconds...

---

At the end, to be fair, if your prefer this new version go ahead. But yeah, it's meaningless optimization.

ozh: OK, I'll remove this then, thanks for your feedback. What you describe is what I thought, just wanted your input to be sure.

ozh:
More random inconsistency with `phpunit --repeat` : sometimes I have a failure on `\tests\tests\utilities\get_request.php:111`

```
GetRequest_Tests::test_get_request with data set #32 ('MiXeD', 'http://SHO.rt/', '/MiXeD')
Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
-'MiXeD'
+'mixed'
```

Yet this test is absolutely DB or option or whatever independent. Djeezus.

ozh:
Meeeeh. I think I've found.

LeoColomb:
What a suspens! What was the cause?

ozh:
@LeoColomb testing the next shorturl id was failing when the previously loaded plugin was "random-shorturl". Captain Obvious to the rescue. The whole "let's load a real plugin to test the plugin API" was so dumb it's appalling in retrospective :)

LeoColomb:
Nice one! 🎉 👍

Issue #2634, opened 2020-03-31
session timeout value is high


sathish1981:
@ozh 
We have ran a security testing against the tool and the team reports like below

Current Session expiry time is 7 days
Generates Static Session IDs
Failed to invalidate session after logout
Frameable response (potential Clickjacking)

Do you have any idea, how to solve these?

ozh:
You can use shorter values if you need, see https://github.com/YOURLS/YOURLS/blob/master/includes/Config/Config.php#L188-L194
Add custom values in your config.php. YOURLS_COOKIE_LIFE is the duration of the cookie login, YOURLS_NONCE_LIFE defines for how long a link in the admin area is valid

Not sure what "Generates Static Session IDs" means (is this a do? a don't? we don't use sessions in YOURLS)

Same for "Failed to invalidate session after logout" : when you log out, auth cookie is killed, see https://github.com/YOURLS/YOURLS/blob/master/includes/functions-auth.php#L19

sathish1981:
@ozh 
Generates Static Session IDs : It means the signature token generated is not dynamic. Always it has the same random Alpha number string even after multiple login/logouts. Also the token is static for each user. Probably this was most concerning our security team to proceed further.

ozh:
Indeed, the "secret signature token" isn't tied to the cookie login. You can invalid it and regenerate a new one with editing YOURLS_COOKIEKEY.

This said, this secret signature isn't modifiable by a plugin (for instance you could have a custom plugin to generate time limite auth signatures)

I'm going to fix this, this is a very valid point.

sathish1981:
@ozh - Thanks for the reply
I have clear understanding on the issue now, let me explain it

Invalidating Cookie Session:
The real problem is, once logout the yourls session, the cookies getting deleted from browser(client side) but not in the server. The value stored in the cookie can be re-applied to authenticate like the same user. To solve this, we need to remove the existing value stored in server and accept the new dynamic value generated by code.

Static cookie Session Value: 
The value of the cookie is always same for each user it never change. So when you are reusing the same value, people can impersonate as the same user without password. once we were able to different cookie values for each login attempt then this issue would be resolved

These above 2 are the major call outs which is more vulnerable for exploitation.

I completely agree on the signature token is not related to this. 

Thanks,
Sathish

sathish1981:
@ozh 
More or less the mentioned points are interrelated. Do you think with any simple hack this can be overcome? 
Do you have any suggestions for me?

[...]

PR #2638, opened 2020-04-03
Enhance cookie functions

ozh:
Address several issues raised by #2634

### DIFF
--- a/includes/Admin/Logger.php
+++ b/includes/Admin/Logger.php
@@ -35,7 +35,6 @@ class Logger extends \Aura\Sql\Profiler {
      * @param string $message
      */
     public function log($message) {
-        yourls_do_action('debug_log', $message);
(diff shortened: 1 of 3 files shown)


ozh:
This isn't done yet, this is a branch in progress. This said, I don't know exactly how far I'll go, as some of your concerns may be specific to your use case, but at the very least I want them to be addressable via plugin (which isn't possible currently in master branch, hence this one)

There's no stripping of host name I'm aware of, at least in this branch. I suggest try the current master branch an open an issue if the problem persists.

sathish1981:
@ozh 
OKay Sure. What I really did was, just copied your modified functions-auth.php and functions.php. Then replaced in my environment files. By that way I checked and updated my comments. While checking I faced the domain name stripping after authentication. 
Sorry if that was the wrong way to test it.

PR #2644, opened 2020-04-12
Accept timestamped signature with arbitrary hash

ozh:
Improves #2581


ozh:
Documentation updated: https://github.com/YOURLS/YOURLS/wiki/PasswordlessAPI#use-other-hash-algorithms-than-md5

PR #2649, opened 2020-04-17
Use SVG logo in lieu of the old PNG one

ozh:
Ahah. I love this expression, "in lieu of" :)
Also, remove those social buttons. I think they're useless and ugly.


ozh:
Rather trivial html/css tweak, seems to work fine for me on current Win10 FF, Chrome & Edge

Issue #2651, opened 2020-04-18
Double slash (/) in urls


sketchub-official:
I tried YOURLS, and everylink which is produced is like http://example.com//123 while it should be http://example.com/123 .

sketchub-official:
Sorry, it is solved, in 'YOURLS_SITE' i placed https://example.com/ instead https://example.com

ozh:
I thought this was fixed anyway. What version of YOURLS are you using?

sketchub-official:
Im using latest one. Yes it is fixed, in YOURLS_SITE i accidentally placed "example.com/"instead "example.com".

ozh:
No, I mean, I thought entering "yoursite.com/" instead of "yoursite.com" wasn't a problem any longer

PR #2653, opened 2020-04-19
Always return trimmed (and filtered) YOURLS_SITE

ozh:
Should fix for good issues like #2651


ozh:
(this `functions.php` is getting uglier and uglier. I'd like to split it in a couple other files but I cannot think of proper file names and content 😕)

Issue #2654, opened 2020-04-19
`functions.php` is messy. Ideas for splitting?


ozh:
The file has 2500 lines and more than 100 functions.

I'd like to split it in a couple other files (`functions-something.php`) but cannot think of a proper, relevant and illustrative `-something` ... Any ideas?

(similarly, any ideas to split other large files into relevant tinier pieces welcome)

paulgoodchild:
I guess you could gather related functions into categories, like:

- filters (e.g. `yourls_return_true`, `yourls_return_false`)
- keywords (e.g. `yourls_keyword_is_reserved`, `yourls_keyword_is_free`)
- request state (e.g. `yourls_is_infos`,`yourls_is_admin`,`yourls_is_API`)

paulgoodchild:
Would you like me to make a start on cleaning some of this with the suggestions above and offer a pull request?

ozh:
Maybe the most practical thing first would be to simply have a document with file names and function names, rather than a whole codebase to mess with ? Something like http://collabedit.com/prpbd ? Thoughts?

paulgoodchild:
Sure... you mean to just plan it out this way?
I've updated that doc and moved a few functions with:
# functions-debug.php
# functions-filters.php
# functions-options.php
# functions-state.php

Of course, names can be adjusted, it's just to start some sort of grouping.

ozh:
Not looking bad. I'll need to have this running a couple days in my head eventually but it's starting to look nicer :)

Maybe another grouping could be "stuff relative to short URL" (yourls_add_new_link, yourls_*_keyword_*...) and "stuff relative to how we mangle links or URL" (yourls_admin_url, yourls_site...) but my brain crashed at the naming problem (the whole app is about links, so for instance having functions-links.php or functions-urls.php sounded dumb to me)

paulgoodchild:
My view on that is it does no harm to rearrange and organise it, and if you change your mind later on for the naming the convention, it's a simple to adjust it. Best to start with something small and build from there, or you'll never move it.

Separately, another place to clean a few things up which we can do in this reorg, is to use things like shorthand array notation and adjust simple functions like this to be cleaner and less verbose:

```php
function yourls_is_API() {
    $return = defined( 'YOURLS_API' ) && YOURLS_API == true;
    return yourls_apply_filter( 'is_API', $return );
}
```
to
```php
function yourls_is_API() {
[...]

PR #2667, opened 2020-04-30
Move some stuff from functions.php to smaller function files

ozh:
Expanding on @paulgoodchild suggestions from #2654

* [x] `functions-options.php`
* [x] `functions-debug.php`
* [x] `functions-shorturls.php` for stuff related to adding/editing short URLs
* [x] `functions-links.php` for stuff related to how we manipulate links

TBD:
* [x] functions-filters.php, because it's to be used by both filters *and actions*, not just filters, so I don't like the file name. Either `functions-hook-helpers.php` maybe, or just putting these functions in `functions-plugins.php`?


ozh:
Also maybe:
* [ ] there are a couple functions in `functions.php` that make or relate to HTTP requests. Either move them in `functions-http.php`, or move them in an aptly named new file ?

ozh:
Now every file has few enough function declaration so that the whole list fits vertically on my screen, that's better and fine with me :)

PR #2670, opened 2020-05-01
Begin functions cleaning, starting with GEO.

paulgoodchild:
- also adds support for CloudFlare headers to shortcircuit need for GeoIP database.


paulgoodchild:
The coding styles I use in projects is generally [aligned with PSR-12](https://www.php-fig.org/psr/psr-12/)

paulgoodchild:
On that failure, I'd say perhaps better to remove the "assertFalse" and instead ensure that the function `yourls_geo_get_flag()` will return the default "empty" flag on failure, instead of `false`

ozh:
Nice touch about the Cloudfare stuff, I wasn't aware of this.
You're using even more white space around brackets and parenthesis and stuff than me 🤣

ozh:
(@paulgoodchild just say when you think you're done with a branch and I'll merge)

paulgoodchild:
@ozh I figure this can be ongoing and you can probably merge as we go if you're happy with what you see.

I do like myself a bit of whitespace! :D  It just makes everything a bit more readable I find :)

ozh:
Can I merge as we go ? I think this will close the PR
On the other hand I think it's easier to merge frequently little chunks, this avoids conflits
If this is cool with you I merge here and you continue with other files in another PR ?

paulgoodchild:
yep, happy to create another PR as I go.. no point holding off

PR #2672, opened 2020-05-02
Allow 255 chars for the user agent, as this is the column size in the DB

mibe:
Allow 255 chars for the user agent, as this is the column size in the DB. The limit of 255 is also enforced in the calling function [yourls_log_redirect](https://github.com/YOURLS/YOURLS/blob/c6d9710c6284c865835659d5a5a70bd4484a1089/includes/functions.php#L405).


ozh:
Indeed :)

PR #2673, opened 2020-05-02
New plugin hook for manipulating the HTTP referrer

mibe:
This allows plugins to change the HTTP referrer.

The code adds a new function, ```yourls_get_referrer``` which is called by ```yourls_log_redirect```. A new hook is added, ```get_referrer```, so that plugins can change the referrer information.

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -207,10 +207,7 @@ function yourls_get_user_agent() {
  * @return string               HTTP Referrer or 'direct' if no refferer was provided
  */
 function yourls_get_referrer() {
-    if ( !isset( $_SERVER['HTTP_REFERER'] ) )
-        return 'direct';
-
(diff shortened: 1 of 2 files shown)


ozh:
Sure, why not. This said, the filter should always be the last thing before return so plugins can always manipulate the value. Here, with no referrer set, a plugin cannot catch the `'direct'` return value. The logic needs to be changed to 
- do checks
- assign value
- only one return in the function, filtered

mibe:
Alright, I've changed the code so that the plugin is also called when no referrer was provided.

BTW: I used ```yourls_get_user_agent``` as a template. If there is no user agent set, it will return without calling the plugin. Is this a bug then?

ozh:
Not a "bug" but definitely something I could have been more careful about :) Fancy fixing that yourls_get_user_agent() as well while you're at it?

ozh:
Thanks!

PR #2679, opened 2020-05-06
Add update YOURLS logo

JayHoltslander:
@ozh requested the update to the existing svg to change the chain links.
Added icon version .svg as well.


dgw:
Weird how this changes the preview size of the logo to be so much smaller 😕

ozh:
@JayHoltslander neat!
Since it's "scalable vector", is there any impact modifying the width and  height of the SVG? I see you changed from 9081 to 512 : does that modify only the "default" view, or will this change the way browser will display the image for instance?

JayHoltslander:
Yeah it's scaleable up or down. But at least at 100% it's a more reasonable size. I'm not sure why I left the first version so big before committing it. I had it sized enormously for the purposes of tracing over the original raster graphic.

ozh:
Questions for the team, @JayHoltslander @dgw @LeoColomb : 

## Logo update

Current SVG logo, based on the crappy low res paintbrush I did 10 years ago is now this:


The "full" logo I asked @JayHoltslander to SVGize would be this (updated chain links, I think it was @LeoColomb's idea at the time, more modern "flat" look)


**Question**: everybody thinks it's better with the updated version ?

## Icon

The idea in this updated logo was to use the tiny chain link as a standalone icon where needed, again to update the crappy 16x16 pixel I did. This icon @JayHoltslander also redesigned as follows:


which slightly differs from the tiny link from the updated logo:
image

**Question**: which one of these 2 shapes do you peeps prefer ?

dgw:
Personally, I don't read the new chain as a chain. It looks like two wheels joined at the hubs, which doesn't quite scream "link". :/

ozh:
> 
> 
> Personally, I don't read the new chain as a chain. It looks like two wheels joined at the hubs, which doesn't quite scream "link". :/

Hmm, I admit I've had that "wheel" feeling years ago but I got over it. You think there are peeps landing on http://blog.yourls.org/ thinking it's something about futuristic cars or bicycles? :)

dgw:
Heh. Maybe use the version with gaps around the horizontal link? It's "more" chain-y. 😝

JayHoltslander:
Bear in mind that you're looking at the icon on white.
See: https://codepen.io/j_holtslander/pen/NWGyyZO

dgw:
That's the gap version I was talking about!

I can think of a lot of cases where it would be displayed on white, the default background color for web pages.

JayHoltslander:
Could just use an existing icon for any instances of small icon  <shrug>
Eg: https://www.materialui.co/icon/link
Or: https://fontawesome.com/icons/link?style=duotone

ozh:
Nice touch with the white border around both logos, I hadn't noticed.

I'd rather have an image for the icon rather than a font because it'll be easier for other people to use it *when we have some theme API* and anyway we also need it for favicon and real images

So, it would be : 
Logo:

Icon:


Are we all cool with this (I am 😄), or does someone wants to try something completely new ?

dgw:
I'd like it better if the logo used the same link style as the icon. Even more better if the colors all matched 😅

JayHoltslander:
A thought/design version occurred to me.
Will mock it up this evening.

JayHoltslander:
Ok honestly if it were up to me... this is what I would want to go with:
https://codepen.io/j_holtslander/pen/RwWMKpx

ozh:
❤️ ❤️ ❤️ !!! Yes ! This !

dgw:
^ what he said 😁

@ozh do you mind about making sure the link icon and the link component of the logo match colors?

ozh:
You mean giving the little chain the same color as the YOURLS letters? I tried but I think the logo gets a bit dull, too "flat"... I think having the chain to "pop", to be "highlighted" a bit is nicer: https://codepen.io/ozh/pen/zYvWEdb
(jesus I wish I was good at css and design :)

What I think makes most sense is: having the logo as @JayHoltslander designed (with YOURLS in lighter blue and the chains in darker blue), and the icon (the last bit of the chain) in lighter blue (which I identify as the "YOURLS color")

Thoughts?

dgw:
@ozh I meant matching the icon to the color of the chains in the logo, using that darker blue for both. But I don't want that any more 😂

I like your top/first example, though I couldn't see the whole codepen on my phone.

JayHoltslander:
Some icon thoughts...

dgw:
Third column for me, color-wise. I think that circular design would be good as a favicon! Round is in these days, haha.

JayHoltslander:
Well technically a logo could appear square or round depending on where it's placed.
You should think of that last screenshot as 4 versions seen in different environments.

ozh:
Very nice job again @JayHoltslander, I really like this! Superb little icon.

I think we're all good: full logo is slick, this little icon is slick. It will be used instead the current favicon.gif, either inline in HTML docs or as a favicon
image
image
image

I think I'll use this to make a full favicon package and plugin (complete with [all the garbage that's needed](https://sympli.io/blog/heres-everything-you-need-to-know-about-favicons-in-2020/) for ios, android and windows stuff).

@JayHoltslander could you clean up the PR here? Remove my 2 poor SVG attempts, add the updated yourls-logo.svg and yourls-icon.svg ?

JayHoltslander:
Will come back to this this weekend.

Aside: I’m a fan of realfavicongenerator.net

LeoColomb:
Wow, great job @JayHoltslander!
Looks promising.

> realfavicongenerator.net

Yes very good reference! You can use it to generate other require icon, but please use standard configuration.

---

@YOURLS/admins Can I ask for a ping during final review before merge? I would appreciate, thanks!

ozh:
ping @JayHoltslander (no rush at all, just making sure you're still alive 😄)

JayHoltslander:
Hi yes. Still alive. Sorry. Super swamped with IRL stuff. Will do as soon as possible.

ozh:
@JayHoltslander should we send help & food your way ? :) (hoping it's just business real life stuff)

ozh:
Ping @JayHoltslander 😺  (Feel free to just send me files over email and I'll take care of things)

ozh:
Happy new year @JayHoltslander ;)

PR #2684, opened 2020-05-09
Timezones

ozh:
Build on @JeremyKeusters's #2633 to

* have timezone management compatible "core" (the functions in function files), but not ship a new core plugin. This way, if the plugin has bugs, or if we overlook a trivial feature, we can quickly update it without needing to push a new YOURLS release that would impact everybody including people not using the core plugin
* have a timezone management plugin, with its own release cycle. Eventually, depending on adoption and feedback, it could become a core plugin like "random keywords" became.


ozh:
Plugin moved to https://github.com/YOURLS/timezones

ozh:
Might be all done, this PR and https://github.com/YOURLS/timezones seem to work fine :)

PR #2689, opened 2020-05-15
Contributors

ozh:
Follow-up to 00a437b35f981312628b7af68495650982cd60b4, which didn't quite worked as expected

Yet another almost-meaningless attempt at giving some people a little credit


ozh:
Nope, didn't work better :)

PR #2690, opened 2020-05-17
Thatfuckingencodingproblem

ozh:
Feedback welcome -- See issue #2691


ozh:
Take that, https://طارق.net/طارق !

(no idea what this domain is about but they sure got some traffic from my dev setup this past few days)

ozh:
Current state of things:
* _mostly_ works (I know one scenario with bookmarklets that isn't perfect, but I can live with this)
* now more torture tests on localhost and https://test.yourls.org/ to try to break things

ozh:
Take that, http://héhé.yourls.org/ !

ozh:
Kind of stuck on something 😕 

This expectedly works in a `test.php` : 
```php
$url = 'http://普遍接受-测试.世界';
var_dump($url === yourls_sanitize_url( $url ));
// true
```

but fails in a unit test : 
```php
$url = 'http://普遍接受-测试.世界';
$this->assertEquals( $url, yourls_sanitize_url($url) );
// Failed asserting that two strings are equal.
// --- Expected
// +++ Actual
// @@ @@
// -'http://普遍接受-测试.世界'
// +'http://普?__接?_?-测试.世界'
```
(doesn't change with `assertSame()` or `assertTrue( $url === yourls_sanitize_url( $url ) )`)

admin/upgrade.php
```diff
@@ -28,8 +28,8 @@
 
 	// From what are we upgrading?
 	if ( isset( $_GET['oldver'] ) && isset( $_GET['oldsql'] ) ) {
-		$oldver = yourls_sanitize_version( $_GET['oldver'] );
-		$oldsql = yourls_sanitize_version( $_GET['oldsql'] );
+		$oldver = (string)( $_GET['oldver'] );
+		$oldsql = (string)( $_GET['oldsql'] );
```
LeoColomb, line 32: Why not using the sanitize function? And not really related to the encoding issue, right?

composer.json
```diff
@@ -25,7 +25,8 @@
         "geoip2/geoip2" : "2.10.0",
         "aura/sql": "~2.",
         "jakeasmith/http_build_url": "1.0.1",
-        "symfony/polyfill-mbstring": "1.15.0"
+        "symfony/polyfill-mbstring": "1.15.0",
+        "symfony/polyfill-intl-idn": "^1.17"
```
LeoColomb, line 29: 👍

includes/Config/Init.php
```diff
@@ -85,14 +85,16 @@ public function __construct(InitDefaults $actions) {
         // Check if need to redirect to install procedure
         if ($actions->redirect_to_install === true) {
             if (!yourls_is_installed() && !yourls_is_installing()) {
-                yourls_redirect( yourls_admin_url('install.php'), 302 );
+                yourls_no_cache_headers();
+                yourls_redirect( yourls_admin_url('install.php'), 307 );
```
LeoColomb, line 89: 👍 for the 307

includes/class-mysql.php
```diff
@@ -30,7 +30,7 @@ function yourls_db_connect() {
         $dbhost = sprintf( '%1$s;port=%2$d', $dbhost, $dbport );
     }
 
-    $charset = yourls_apply_filter( 'db_connect_charset', 'utf8' );
+    $charset = yourls_apply_filter( 'db_connect_charset', 'utf8mb4' );
```
LeoColomb, line 33: Ark, breaking change, right?

includes/functions-upgrade.php
```diff
@@ -27,6 +30,9 @@ function yourls_upgrade( $step, $oldver, $newver, $oldsql, $newsql ) {
 		if( $oldsql < 482 )
 			yourls_upgrade_482();
 
+		if( $oldsql < 505 )
+			yourls_upgrade_to_18();
```
LeoColomb, line 34:
```suggestion
			yourls_upgrade_to_505();
```

For consistency 🙂

includes/functions-upgrade.php
```diff
@@ -5,7 +5,10 @@
  *
  */
 function yourls_upgrade( $step, $oldver, $newver, $oldsql, $newsql ) {
-	// special case for 1.3: the upgrade is a multi step procedure
+
+    yourls_maintenance_mode(true);
```
LeoColomb, line 9: 👍

includes/vendor/composer/autoload_files.php
```diff
@@ -6,6 +6,8 @@
 $baseDir = dirname(dirname($vendorDir));
 
 return array(
-    'b45b351e6b6f7487d819961fef2fda77' => $vendorDir . '/jakeasmith/http_build_url/src/http_build_url.php',
     '0e6d7bf4a5811bfa5cf40c5ccd6fae6a' => $vendorDir . '/symfony/polyfill-mbstring/bootstrap.php',
+    '25072dd6e2470089de65ae7bf11d3109' => $vendorDir . '/symfony/polyfill-php72/bootstrap.php',
```
LeoColomb, line 10: Ark 😉

tests/tests/format/urls.php
```diff
@@ -80,6 +81,9 @@ function list_of_valid_URLs() {
             array( '/absolute' ),
             array( '/Absolute/Path/' ),
             array( '/absolute/path/?omg#also' ),
+            array( 'http://académie-française.fr' ),
+            array( 'http://www.طارق.net/طارق?hello=%2B' ),
+            array( 'http://%d8%b7%d8%a7%d8%b1%d9%82.net/' ), // this is طارق.net, encoded. I _think_ it qualifies as valid
```
LeoColomb, line 86: Well, not sure at all. IDN should keep their official encoding system, here: `http://xn--mgbuq0c.net/`
Best option is to only store `xn--` version of domains.

includes/functions-formatting.php
```diff
@@ -562,7 +565,7 @@ function yourls_esc_url( $url, $context = 'display', $protocols = array() ) {
  * @param string $url URL
  * @return string URL with lowercase scheme and protocol
  */
-function yourls_lowercase_scheme_domain( $url ) {
```
LeoColomb, line 565: High level of breaking change here. 😨
Is the function moved to compat somewhere?

admin/upgrade.php
ozh: Not related to encoding. It was breaking the build because the function doesn't do as advertised. I should take care of it btw.

includes/functions-upgrade.php
ozh: Actually I think this is `yourls_upgrade_482()` that was inconsistent. Much easier to see what has changed from the version number/tag than to git blame and check when that arbitrary number was bumped

includes/class-mysql.php
ozh: Not breaking in the sense of "something that was working previously will stop working", that's what the upgrade function is there for. I think that's a typical case for [minor version](https://semver.org/#spec-item-7) (that includes "patches" changes) hence 1.8

includes/functions-formatting.php
ozh: Good call. I'm adding it to `functions-deprecated.php`.

includes/vendor/composer/autoload_files.php
ozh: Added by symfony/polyfill-intl-idn ...

tests/tests/format/urls.php
ozh: I added this test because during my testing (shorten normally in admin area, use bookmarklet, use `http://sho.rt/` prefix) I ended up with this. Browsers are happy with it...

admin/upgrade.php
ozh: In fact I was using `yourls_sanitize_version()` for YOURLS version, like `1.5.1`, fine, but also for the SQL version, eg `480`, in this case in was not fine (not a "version number" and was returning `''`). Anyway type casting is enough here.

PR #2732, opened 2020-07-26
Update functions-html.php

flortsch:
Fix logout link in admin panel.


ozh:
Why ? With default value `''` (empty string) it redirects to `<root>/admin/` ([source](https://github.com/YOURLS/YOURLS/blob/cab0c150a0ce2b0f8775a029c6b47fa98d69598b/includes/functions-links.php#L155)), which works fine normally

flortsch:
Yeah, but the redirect to `<root>/admin/` did not work for me on an nginx server with the config provided in the wiki. It resulted in a 403 forbidden error.

PR #2744, opened 2020-08-13
Update functions-html.php

ueickmann:
To enable a filter-function to decide separately for each entry wether a user should get the edit/delete/... buttons displayed, the keyword is a necessary context information and should therefore be provided as parameter.


dgw:
@LeoColomb This won't break existing hook functions that expect just one arg, right?

PR #2747, opened 2020-09-08
Allow changing the "/admin/" part of the URI

havill:
Like how WP and many other services allow changing the admin URL, allow the admin directory name to be changed to make it just a little harder for a bot to do a brute force attack at the login area.

Overriding the default "/admin/" is done through the config.php page like other related settings:

```
define( 'YOURLS_ADMIN_DIR', 'MyAdm1nistrationZ0ne' );
```

### DIFF
--- a/includes/functions-links.php
+++ b/includes/functions-links.php
@@ -157,7 +157,8 @@ function yourls_admin_url( $page = '' ) {
     } else {
         $yourls_admin_dir = "admin";
     } 
-    $admin = yourls_get_yourls_site() . '/' . $yourls_admin_dir . '/' . $page;    if( yourls_is_ssl() or yourls_needs_ssl() ) {
+    $admin = yourls_get_yourls_site() . '/' . $yourls_admin_dir . '/' . $page;
+    if( yourls_is_ssl() or yourls_needs_ssl() ) {
         $admin = yourls_set_url_scheme( $admin, 'https' );
     }
     return yourls_apply_filter( 'admin_url', $admin, $page );
--- a/includes/functions-links.php
+++ b/includes/functions-links.php
@@ -152,8 +152,12 @@ function yourls_statlink( $keyword = '' ) {
  *
  */
 function yourls_admin_url( $page = '' ) {
-    $admin = yourls_get_yourls_site() . '/admin/' . $page;
-    if( yourls_is_ssl() or yourls_needs_ssl() ) {
+    if ( defined( 'YOURLS_ADMIN_DIR' ) ) {
+        $yourls_admin_dir = YOURLS_ADMIN_DIR;
+    } else {
+        $yourls_admin_dir = "admin";
+    } 
+    $admin = yourls_get_yourls_site() . '/' . $yourls_admin_dir . '/' . $page;    if( yourls_is_ssl() or yourls_needs_ssl() ) {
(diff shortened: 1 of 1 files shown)


ozh:
Why introduce a new constant when a plugin should be enough?

2 mn hack:

1) rename your `admin` directory to `OMGSECRETURL`
2) create a file `user/cache.php` with the following content : 

```php
<?php
// introduce a new filter early, before plugins are actually loaded
yourls_add_filter( 'admin_url', 'ozh_custom_admin_url' );
function ozh_custom_admin_url($url) {
    return str_replace('/admin/', '/OMGSECRETURL/', $url);
}
```

This little known trick allows a file to be included very early, see https://github.com/YOURLS/YOURLS/blob/1.7.9/includes/Config/Init.php#L60-L63

If anything we'd rather suppress constants that adding new ones, and deprecate them in favor of a plugin solution which is more user friendly.

I admit this 2 mn hack is not really a "plugin", but it's the same approach as editing a config file to add a custom defined constant and renaming a core directory.

This said I'm very open to improvements to make this possible via a real plugin :)

By the way maybe we should rename this `user/cache.php` to something like `user/must-use-plugins.php` and more document it :)

havill:
Nothing against plugins. It's just that in this particular case, the change was particularly low-level with a hard-coded string "/admin/" used in a core function used below the plugin layer, with particular consequences if things go wrong (locked out of administrating the service if things go wrong).

As you demonstrated above, in order to get your hack to work, you needed to do it below/before the plugins are called.

Q: what happens if some plugin needs to know where the admin directory/code is, and that plugin is called before the plugin that "changes" the admin directory location?

Tequenikality (approved):
Having looked through the changes made to functions-links.php, the changes seem (at first glance) perfectly acceptable. I do question the practical uses for the concept but the execution seems valid.

However, I would have expected that there'd be other hard-coded references to /admin within YOURLS core. I haven't seen them and I don't know if there are for sure, but I would think so. Testing would easily reveal conclusively.

dgw:
@Tequenikality The point of `yourls_admin_url()` is so there _won't_ be hard-coded references to `/admin` anywhere else.

Tequenikality:
> @Tequenikality The point of `yourls_admin_url()` is so there _won't_ be hard-coded references to `/admin` anywhere else.

I did say that I wasn't sure - besides, coding a method like that suggests that changing the admin URL was an intended functionality. Why go to the trouble of coding a function like that and coding the rest of core to use it, just to skimp on the functionality ultimately? It doesn't make much sense to my mind. That's another reason why I suggested testing - if it was that simple to implement, why wouldn't it be a stock option? There must be a reason that Ozh skipped on the functionality and just hard-coded the function.

dgw:
> There must be a reason that Ozh skipped on the functionality and just hard-coded the function.

Probably just that nobody asked for this before. The function exists mostly to determine the correct base URL (`YOURLS_SITE`), path, and scheme (HTTP/HTTPS) given the site settings. There's enough variation that I'd hardly call it "hard-coded".

LeoColomb:
@YOURLS/core Shall we close this one as plugin domain?

ozh:
Yep - and let's document somewhere the `user/cache.php` trick (maybe rename this file, even). I shall do this, see  #3045

PR #2754, opened 2020-10-01
The mac client is deprecated

aaryanporwal:
The mac client is deprecated and the link is redirecting to other website (http://timi.im/app/) which looks malicious.


dgw:
Not sure how active Tim is, but @devtimi any objection to removing this link? Is there a new one?

aaryanporwal:
Yeah I don't think they are active, and it doesn't seem good if folks are unknowingly redirected to a malicious website.

aaryanporwal:
Here is a web archived link which still works:
https://web.archive.org/web/20131104052112/http://timi.im/app/
The download file is intact too.
Note that this v0.4 of the app which is the most recent one that had been archived, I ran it on my machine and it surprisingly still works!

devtimi:
I'm sorry, I don't have plans to update the Mac desktop client.

Issue #2714, opened 2020-06-16
Replace `call_user_func()` with variable functions


ozh:
Apparently, `call_user_func()` is slow. References:
* [php.net](https://www.php.net/manual/en/function.call-user-func.php#61553)
* [fab2s/call_user_func](https://github.com/fab2s/call_user_func)
* symfony/symfony#29309

We're using a couple (`functions-l10n.php` and `functions-plugins.php` as of writing) so the idea would be to replace them:
```php
// before : 
call_user_func( $function1 );
call_user_func( $function2, $param );
// after :
$function1();
$function2( $param );
```

While at it, an `is_callable()` check before calling would be nice, to ensure the function is actually callable when it's user defined (plugins) (and make sure the cases of calling something callable and not callable are covered in a unit test)

ozh:
On the contrary and for the record, `call_user_func_array($func, $array)` seems faster than `$func(...$array)` on recent PHP build -- see https://gist.github.com/nikic/6390366

The remark about checking with `is_callable()` is still valid, though

PR #2763, opened 2020-10-11
Replace `call_user_func()` with variable functions

ozh:
Backstory: #2714 

Idea:
```php
// before : 
call_user_func( $function1 );
call_user_func( $function2, $param );
// after :
$function1();
$function2( $param );
```

Also, add checks for `is_callable()` where relevant, and update tests accordingly.

### DIFF
--- a/includes/functions-plugins.php
+++ b/includes/functions-plugins.php
@@ -149,7 +149,7 @@ function yourls_apply_filter( $hook, $value = '' ) {
     reset( $yourls_filters[ $hook ] );
     do {
         foreach ( (array)current( $yourls_filters[ $hook ] ) as $the_ ) {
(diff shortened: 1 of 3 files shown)


ozh:
I added a check, and unit tests, for the callable when a plugin admin page is drawned, see `yourls_plugin_admin_page()`

I didn't add a check for a callable in actions and filters hook -- see [`yourls_apply_filter()`](https://github.com/YOURLS/YOURLS/blob/1.7.9/includes/functions-plugins.php#L145-L157) : I prefer a malformed hook to display a fat ugly error, rather than catching the problem and displaying a nice message. Unit tests are covering this case, see https://github.com/YOURLS/YOURLS/blob/1.7.9/tests/tests/plugins/actions.php#L422-L433

PR #2764, opened 2020-10-12
Unobtrusive message for people peeping in their console

ozh:
Thoughts ?


dgw:
"toggable" message?

joshp23:
this might be interesting if the message could be filtered

ozh:
Yeah, toggable with filter `shunt_js_easter_egg`

Regarding filtering the message itself: well if someone wants to display their own, it has always been possible using action `html_head`. It's not really a new "feature" per se

ozh:
So, I understand this isn't popular. Ok :)

PR #2767, opened 2020-10-14
Issue - 2683 Default shorturls_length in the database if its not already set

### LINKED ISSUE #2683: Random ShortURLs plugin doesn't show default value
Upon first activation it should show the default value instead of a blank input:
image

(Also, not sure the HTML control is what we need here, a simple input text would be simpler. Was I the one to come up with this idea in the original plugin? Don't remember)

zach2825:
Also in the gitignore file, I added an entry for the idea directory

image

Closes #2683

### DIFF
--- a/user/plugins/random-shorturls/plugin.php
+++ b/user/plugins/random-shorturls/plugin.php
@@ -64,7 +64,7 @@ function ozh_random_shorturl_settings_page() {
     $nonce = yourls_create_nonce( 'random_shorturl_settings' );
 
     // when $random_length is false set it to 5 and persist the value in the database
-    if ($random_length === false) {
+    if ( $random_length === false ) {
         $random_length = $_POST[ 'random_length' ] = 5;
         ozh_random_shorturl_settings_update();
     }
--- a/user/plugins/random-shorturls/plugin.php
+++ b/user/plugins/random-shorturls/plugin.php
@@ -61,7 +61,7 @@ function ozh_random_shorturl_settings_page() {
     }
 
     $random_length = yourls_get_option('random_shorturls_length');
-    $nonce         = yourls_create_nonce('random_shorturl_settings');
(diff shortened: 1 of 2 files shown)


user/plugins/random-shorturls/plugin.php
```diff
@@ -61,7 +61,13 @@ function ozh_random_shorturl_settings_page() {
     }
 
     $random_length = yourls_get_option('random_shorturls_length');
-    $nonce = yourls_create_nonce( 'random_shorturl_settings' );
+    $nonce         = yourls_create_nonce('random_shorturl_settings');
```
dgw, line 64: Isn't this also auto-formatting?

zach2825: Thank you, keeps my honest. I did a few tests to make sure my ide will not auto-format anymore.

dgw (approved):
LGTM, assuming it matches @ozh's intent in #2683. 😸

Merger: don't forget to squash!

ozh:
No need for the `if` block : `yourls_get_option()` accepts a default parameter when the value isn't set, see line 37 of the plugin. Less code, and saves a DB query storing the value.

zach2825:
Then the value would only show on the form it does not mean it would exist in the database.

zach2825:
As it's written now would cause a bug.

ozh:
Why would it cause a bug? On the 2 occurrences of `yourls_get_option()`, we now have a default value.

"Fetching a value, and if not exists declare the value is 5" is equivalent to "Fetching a value, if not exists declare it is 5 and store it". You just skip the "store it" step.

zach2825:
You're right, in YOURLS it's required to get options using the get_option method. I was thinking if people get the option directly from the database, which is a non-scenario.

Thank you

PR #2771, opened 2020-10-17
Update PHP reqs to 7.2+

ozh:
This will be needed for #2770


ozh:
(I don't understand why tests pass for PHP 7.0 and 7.1 🤣)

LeoColomb:
(@ozh Because the composer lock file has not been update 😉)

ozh:
composer.lock is synced, and I don't think it has something to do. I'm surprised to YOURLS installs on 7.0 and 7.1 at all, given the update in `yourls_check_php_version()` and [these lines](https://github.com/YOURLS/YOURLS/blob/master/admin/install.php#L22-L25)

LeoColomb:
@ozh Ah indeed, the lock file is not used during this installation, my bad.
Are you sure the function is actually run during the tests?

Anyway, I think it's a good idea to release a new version before raising requirement to 7.2.

ozh:
Superseded by #2824

PR #2788, opened 2020-11-08
Update clipboard.js to 2.0.6

whalehub:
The only breaking change since v1.6.1 was the change to the constructor from `new Clipboard()` to `new ClipboardJS()` in [v2.0.0](https://github.com/zenorocha/clipboard.js/releases/tag/v2.0.0).


ozh:
Thanks! In the off chance that someone is using `new Clipboard()` in their front end or plugin, could we have a compatible constructor?

whalehub:
@ozh Are you referring to plugins/frontends which use the `clipboard.js` file that ships with YourLS rather than bundling/importing their own version?

ozh:
Yes

whalehub:
@ozh I just finished manually checking every repository that is listed [here](https://github.com/YOURLS/awesome-yourls) and surprisingly wasn't able to find one which does that.

These are all the plugins and frontends which use the `new Clipboard()` constructor and how they import the required dependency:

**Plugins**
- [Expiry](https://github.com/joshp23/YOURLS-Expiry): Imports the [clipboard function](https://github.com/joshp23/YOURLS-Expiry/blob/master/expiry/sample-public-front-page.php.txt#L74) directly from `share.js`.

**Frontends**
- [∞²](https://github.com/tomslominski/infinity-squared): Uses its [own copy](https://github.com/tomslominski/infinity-squared/blob/master/footer.php#L31) of `clipboard.js`.
- [Sleeky](https://github.com/Flynntes/Sleeky): Imports `clipboard.js` [from a CDN](https://github.com/Flynntes/Sleeky/blob/master/sleeky-frontend/frontend/header.php#L57).

This pull request can be safely merged without breaking any plugins or frontends.

ozh:
Neat work! Thanks for taking the time to go through this. In the end I think the correct way would be to use the `share.js` init function like the plugin does.

PR #2791, opened 2020-11-18
URL encoded to base64

Drizt87:
Recomended to skip "Possible Remote File Inclusion (RFI) Attack: Off-Domain Reference/Link" on some plesk servers with modsecurity enabled.


ozh:
I don't think we'll merge for everyone, and the dozens of thousands of YOURLS installs, something that's intended to accomodate "some plesk servers with mod_sec".

Drizt87:
This change only improves functionality, and does not affect the operation of the application

PR #2804, opened 2020-12-18
Fix encoding when editing existing shortcut

cdchase:
Previously URL encoded links are decoded before adding **New** shortcuts in `yourls_add_new_link()`.
This updates `yourls_edit_link()` to do the same for **Edits** of existing shortcut.


ozh:
I think this is fixed is current master

PR #2822, opened 2021-01-31
SVG icons

simon-eller:
I uploaded the SVG icons and replaced the png and gif files with them.


ozh:
This could be a nice improvement. At the moments I see two things to tweak : 

CSS or icons : action icons are a bit off in their square, like aligned on the right and the sorting filter icons overlap with the text in the table header cells (FF on Win10)
image

The YOURLS icon isn't what it should hopefully be in the (near?) future if we ever hear back from @JayHoltslander :) See #2679

simon-eller:
I fixed it, now it looks like this.
grafik

ozh:
@LeoColomb @dgw thoughts ?

LeoColomb:
Looks good based on the screenshot, but should be tested on every main browser (Chrome, Firefox, ideally IE 11).

ozh:
@simon-eller on my machine result is identical in Edge, Chrome and FF and not different than my 1st screenshot 😕

ozh:
@simon-eller Never mind, I'm nitpicking on you for something that's already the case currently, my bad :)

I think we cannot just use `background-image: url(the icon)`, as it makes it impossible to ensure a proper left margin for the icon. Or the cell header (eg `Clicks`) should be wrapped in its own `<span>` with a right padding.

PR #2823, opened 2021-01-31
Update readme.html

LinuxStans:
Added a new detailed tutorial for LEMP+Let's Encrypt


ozh:
Neat.

In the future I think we should probably move all this to YOURLS/awesome as stated [here](https://github.com/YOURLS/awesome-yourls/issues/59)

ozh (approved):

PR #2828, opened 2021-02-02
Move to GitHub Actions

LeoColomb:

PR #2838, opened 2021-02-15
Keep PHP nightly runs experimental in CI

LeoColomb:
Ref https://github.com/YOURLS/YOURLS/pull/2828#issuecomment-779134121

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,12 +9,14 @@ jobs:
   test:
     name: PHP ${{ matrix.php }}
     runs-on: ubuntu-latest
+    continue-on-error: ${{ matrix.experimental }}
 
     strategy:
       fail-fast: false
       matrix:
         php: ['8.0', '8.1']
         phpunit: ['latest']
+        experimental: [false]
         include:
         - php: '7.2'
           phpunit: '8.5.13'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php: ['8.0', '8.1']
+        php: ['8.0']
         phpunit: ['latest']
         experimental: [false]
         include:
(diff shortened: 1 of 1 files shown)


LeoColomb:
Hmm, weird.

LeoColomb:
For the ref, directly extracted from the doc: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idcontinue-on-error

dgw:
Seems like it would ideally include only one PHP 8.1 check, not two? (Well, four here because it's running both `push` and `pull_request`, which is also excessive but unrelated to the matrix.)
image

LeoColomb:
@dgw Yes, but I don't get how it reads the two times 8.1. And both are reported as error...

.github/workflows/ci.yml
```diff
@@ -9,19 +9,24 @@ jobs:
   test:
     name: PHP ${{ matrix.php }}
     runs-on: ubuntu-latest
+    continue-on-error: ${{ matrix.experimental }}
 
     strategy:
       fail-fast: false
       matrix:
         php: ['8.0', '8.1']
```
dgw, line 17: @LeoColomb What if instead of this, you just added another item to the list under `include:` for 8.0? Maybe that would get rid of the duplicate 8.1 run.

As for `continue-on-error`, I'm not convinced that that actually prevents reporting the overall run as failed; it just prevents canceling other jobs in the run because that one failed. I think. The docs are unspecific, and reading through https://github.com/actions/toolkit/issues/399 didn't really help either.

LeoColomb: Good catch, it was just in front of me. Thanks!

LeoColomb: > I'm not convinced that that actually prevents reporting the overall run as failed

Yes, seems quite a compromised feature 😭 
I'm thinking about just disabling 8.1 for now...

ozh:
Thanks for investigating :)

PR #2933, opened 2021-05-11
bugfix/db upgrade 506

boedah:
Hi!

This PR solves two bugs:

* there was a duplicate array key `short URL type`
* if the DB server has no dynamic row format (see https://mariadb.com/kb/en/innodb-dynamic-row-format/ ), changing the length of the keyword column to 100 (which I guess was done because of maximum key length) after the table conversion resulted in a "key was too long; max key length is 767 bytes"

### DIFF
--- a/includes/functions-upgrade.php
+++ b/includes/functions-upgrade.php
@@ -98,12 +98,12 @@ function yourls_upgrade_to_506() {
     echo "<p>Updating DB. Please wait...</p>";
 
     $queries = array(
-        'database charset'  => sprintf('ALTER DATABASE `%s` CHARACTER SET = utf8mb4 COLLATE = utf8mb4_unicode_ci;', YOURLS_DB_NAME),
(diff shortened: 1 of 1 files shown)


ozh:
Nice catch about the duplicated array key ! My bad, I feel ashamed :)

I don't get the second point. Yes, I've changed the length of `keyword` to `VARCHAR(100)` because of that "key was too long" problem that was [reported numerous times](https://github.com/YOURLS/YOURLS/issues?q=is%3Aissue+%22max+key+length+is+767+bytes%22). AFAIK the problem is solved, could you elaborate more on what change in your PR solves what other problem?

ozh:
Doh. Makes sense... I'm dumb :)

boedah:
As someone who suffered from MySQL row / key length restrictions in various projects for a long time, I can relate :D

boedah:
@ozh IMHO you also could release YOURLS v1.8.2, to make sure the update does work for even more people... :)

PR #3005, opened 2021-07-23
Added default option in db_connect_driver_option to disable SSL Verification by default.

icsy7867:
This will allow the PDO driver to not verify the certificates for MySQL connections using SSL.

In my testing, I have found that this does not affect connection to non-SSL MySQL Databases, but does allow SSL to successfully connect by default with the need for any other additional settings or files (Such as a db.php file as discussed here: https://github.com/YOURLS/YOURLS/discussions/3003 ).

Thus, this should allow for more functionality without hindering or affecting current functionality.

### DIFF
--- a/includes/class-mysql.php
+++ b/includes/class-mysql.php
@@ -51,7 +51,7 @@ function yourls_db_connect() {
      * The driver options are passed to the PDO constructor, eg array(PDO::ATTR_ERRMODE => PDO::ERRMODE_EXCEPTION)
      * The attribute options are then set in a foreach($attr as $k=>$v){$db->setAttribute($k, $v)} loop
      */
-    $driver_options = yourls_apply_filter( 'db_connect_driver_option', [] ); // driver options as key-value pairs
+    $driver_options = yourls_apply_filter( 'db_connect_driver_option', [PDO::MYSQL_ATTR_SSL_CA => false,PDO::MYSQL_ATTR_SSL_VERIFY_SERVER_CERT => false,] ); // driver options as key-value pairs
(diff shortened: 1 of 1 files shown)


LeoColomb (changes requested):
Thanks for opening this pull request @icsy7867.

> but does allow SSL to successfully connect by default with the need for any other additional settings

This also mean the default configuration is less secure, and can allow untrusted connections. 
Also the maintenance cost of adding PDO-MySQL options might be high and can break other/future updates or drivers. It has to be reviewed with care.

icsy7867:
> Thanks for opening this pull request @icsy7867.
> 
> > but does allow SSL to successfully connect by default with the need for any other additional settings
> 
> This also mean the default configuration is less secure, and can allow untrusted connections.
> Also the maintenance cost of adding PDO-MySQL options might be high and can break other/future updates or drivers. It has to be reviewed with care.

This can easily be changed to use a ca.crt instead.

https://github.com/YOURLS/docker-yourls/pull/99/files/e0dee0740e56ae9980ad4357a6017f28322158be#diff-6c077833b6fe8848c6ad2e42736d2bcb1944f624c0118faefb6141be9e528b8c

LeoColomb:
@icsy7867 Indeed, this can be changed easily with the `db.php` file.  
The thing is it's already the case without the suggested additions, right?

Which leads to my comment above, options addition with lower security _per default_ has to be reviewed with care. 🙂

ozh:
There's a filter here that's here exactly to do what this PR would enforce for everyone. 

```php
$driver_options = yourls_apply_filter( 'db_connect_driver_option', [] );
```

If you need to add custom settings to your connection, make a simple plugin and use that filter ?

Issue #3038, opened 2021-08-30
Prevent YOURLS from displaying encryption warning when password is provided via environment variables


bryanpedini:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/YOURLS/.github/blob/main/CODE_OF_CONDUCT.md).

### Submission validity

- [X] This is not a personal support request, that should be posted on [discussions community](https://github.com/YOURLS/YOURLS/discussions).
- [X] I checked [current issues](https://github.com/YOURLS/YOURLS/issues) and this request isn't a duplicate of an existing issue, opened or closed.

### Summary

Since YOURLS 1.7 as per documentation, the software automatically encrypts passwords.
This is not true (of course) for the user provided by environment variables (e.g. Docker).

### Motivation

Just an annoying little error that does not cause any harm but still, it's an error which is actually not true.

### Describe the solution

Check if the array contains only one element (the usual case when providing YOURLS_USER via environment variable) and if true check if the YOURLS_USER and YOURLS_PASS variables are set from environment, in case both checks are true do not even bother to call the function that tries to encrypt the password.

Maybe, just a quick, dirty solution (first that came in my mind, feel free to propose / do something else).
Thanks!

ozh:
A valid point, thanks for raising. Will look into this.

bryanpedini:
Maybe I can refresh some PHP memories and take a look into creating a PR for this if it's welcome (I don't code PHP from quite a while, since my PHP 8.0 framework that I built).....

ozh:
If you can look into this that would be great, indeed!

How I see things after quick toughts : 
- function [`yourls_hash_passwords_now()`](https://github.com/YOURLS/YOURLS/blob/7e06aa1068fb6481834ca95017b4f206d39c2c30/includes/functions-auth.php#L170) is best left unmodified. Its goal is to hash, not to decide if it should hash.
- the test for an env var would be then introduced at [includes/auth.php#L44](https://github.com/YOURLS/YOURLS/blob/7e06aa1068fb6481834ca95017b4f206d39c2c30/includes/auth.php#L44) with a new function like `yourls_user_is_from_env()` or something like this
- a unit test for this new function would be neat

Thoughts ?

bryanpedini:
> Thoughts ?

Nothing concrete yet, I still haven't touched the codebase nor have I looked at how things are structured so that I can formulate a serious proposal about how to implement this.

PR #3040, opened 2021-08-31
Prevent YOURLS from displaying encryption warning when password is provided via environment variables

bryanpedini:
Perhaps a fix for #3038.
Maybe, just a quick, dirty solution (first that came in my mind, feel free to propose / do something else).

EDIT: to be clear, I have not tested this yet, I based my PR on top of documentation, code comments and digging around a little. Using Træfik+Docker in production w/ the official image, I have really no easy way to test this out by myself. Perhaps a unit test could do the work for us.

### DIFF
--- a/includes/auth.php
+++ b/includes/auth.php
@@ -41,26 +41,23 @@ if ( isset( $_GET['dismiss'] ) && $_GET['dismiss'] == 'hasherror' ) {
 } else {
 
 	// Encrypt passwords that are clear text
-	if ( !defined( 'YOURLS_NO_HASH_PASSWORD' ) && yourls_has_cleartext_passwords() ) {
-		// Check if user is provided from environment variables, if true do not bother to encrypt passwords
-		if ( !yourls_check_user_from_env() ) {
-			$hash = yourls_hash_passwords_now( YOURLS_CONFIGFILE );
-			if ( $hash === true ) {
-				// Hashing succesful. Remove flag from DB if any.
-				if( yourls_get_option( 'defer_hashing_error' ) )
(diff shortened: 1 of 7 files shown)


bryanpedini:
I added a wrapper if to be explicit about the checks during the authentication / encryption phases.
The function could have been easily slipped into the outer if but as said above it would be less clear and there would have been no space for the explanatory comment (if necessary).

Otherwise feel free to just remove the wrapper and merge the two checks toghether :)

bryanpedini:
Damn it, thank you for the CI workflows!! I forgot one closing bracket.

Fixed in last commit.

ozh:
Thanks !

ozh:
About `count($yourls_user_passwords) == 1` : I think it's bound to fail at some point. You could totally have a standard YOURLS setup with multiple user/pwd in the config, yet also use another auth mechanism using env vars.

Isn't "username and password are provided by environment" the only red flag we need to care about ?

bryanpedini:
> Isn't "username and password are provided by environment" the only red flag we need to care about ?

Sure... should I just remove it then?

ozh:
I'm refactoring things from your PR, and adding some unit tests, no worries

bryanpedini:
Sure, you have edit access, feel free to edit my branch before merging the PR :)
Thank you for the consideration.

ozh:
Thanks @bryanpedini :) I hope this will fix the original issue :-P (it should)

bryanpedini:
Thank you for giving me the opportunity to contribute! (it's not obvious, let me tell you...)

PR #3090, opened 2021-10-18
fix: Changes to resolve #2688

### LINKED ISSUE #2688: Requesting sho.rt/http://somelink while being logged out displays a notice
It works but displays an incorrect notice

Scenario:
* log out
* access http://sho.rt/http://somelink
* as expected the login form is displayed
* log in
* the link is added (it's not added til you log) but there is a notice about the long link already existing in the DB

If logged prior to shortening, works as expected

dan-r:
This resolves the issues caused in the issue #2688.

The changes to resolve this issue are minor as follows:

- Set the `$user` variable passed to the `yourls_verify_nonce `method to '-1' (unauthorised). There seems to be an issue where the original nonce is created prior to the user logging in, and then verified after the fact. This change ensures that throughout the script execution, the same nonce is being verified
- Add an `exit()` after a redirect in `yourls_is_valid_user()`. Without this, the redirect causes the URL to be added to the database twice (or fail the second time). 

I've tested these changes on my local instance and see no issues.

### DIFF
--- a/includes/functions-auth.php
+++ b/includes/functions-auth.php
@@ -99,6 +99,7 @@ function yourls_is_valid_user() {
 			// Login form : redirect to requested URL to avoid re-submitting the login form on page reload
 			if( isset( $_REQUEST['username'] ) && isset( $_REQUEST['password'] ) && isset( $_SERVER['REQUEST_URI'] ) ) {
 				yourls_redirect( yourls_sanitize_url_safe($_SERVER['REQUEST_URI']) );
+				exit(); 
 			}
 		}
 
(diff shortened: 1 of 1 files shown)


dan-r:
My pleasure - I've been using YOURLS on and off for many years so glad to be able to contribute!

I've replaced the `exit()` with a `return` and the bug still appears fixed.

I haven't had a chance to get the testing setup in my local env yet, but will (hopefully) get them working later on.

dan-r:
@ozh sorry to bump but was there anything else that needed changing with this, or any tests that needed updating as a result? Thanks!

ozh:
Hey, no offense, you're right bumping this, it was sitting on my todo list. Thanks for this PR!

ozh:
Reverted for now, the fix was breaking the tests and after a couple hours on this I decided I prefered a minor UI bug than a whole broken test suite. I'm reopening #2688

PR #3230, opened 2022-02-21
Bump to php 8.1 and drop 7.2, 7.3

LeoColomb:


LeoColomb:
Facing https://github.com/auraphp/Aura.Sql/issues/194, I might consider rolling back Aura.SQL to v3

LeoColomb:
Rolling back to Aura.Sql v3, which is using deprecated functions for 8.1 and contains some very old pieces of PHP (compat 5.3).
Aura.Sql v5 requires PHP 8.1 min (!), and v4 type update might never come according to upstream issue https://github.com/auraphp/Aura.Sql/issues/194.

All the rest has been quite easy to upgrade! ✨
* Supporting PHP 7.4 -> 8.1
* Enforcing minimum php platform for composer
* Deprecated functions for YDB dropped as they can conflict with Aura.Sql

ozh (approved):
Nice work !

ozh:
Huge thanks :)

Regarding the warnings sent by [ozh/phpass](https://github.com/ozh/phpass/issues/5)  : we can simply stop using ozh/phpass since it was needed before 5.5 or 5.6 when PHP `password_*` functions were not available, see 1434
I'll do that.

harikt:
I have been here for this was referenced in Aura. If you guys are using 8.1, can't you switch to aura/sql 5 ? Or is there any compatibility issues you ran ?

Good to see and thanks for using Aura.

ozh:
We're using 7.4 - 8.1

harikt:
Then we probably have no way because union type is only available in 8.0.

LeoColomb:
@harikt Well, the possible workaround would be to remove incompatible return types on version 4.
5 is not an option right now indeed, but I guess 4 was targeting this exact in-between situation, right?

harikt:
@LeoColomb yes. 4 is targetting 7.4 - 8x . If you have a fix, please do send.

harikt:
Hey @LeoColomb ,

I was trying to help here. I am sorry to hear. I have limited time and wanted to finish things early as possible, so I can spend my valuable time with my family.

Regarding closing the issue https://github.com/auraphp/Aura.Sql/pull/208 you sent. I have not closed it. That PR seems closed by you. Also the changes in that cannot be pushed for it is removing the type hints which itself is a BC break.

Anyway if you feel I am disrespectful you can keep that view. But I was trying to help the project and people who use it.

koriym:
Hi @LeoColomb
I am a contributor of Aura project as well as @harikt.

He's doing well for the project with his valuable time and I've never seen him be disrespectful to others. Let's be tolerant of others, even if things don't go well or there are misunderstandings. 🙏

LeoColomb:
Ok ok, let's continue the discussion there.

Issue #2538, opened 2019-07-30
Racing condition results in error message as response but still http 200


schapsl:
**Describe the bug**
When putting load on yourls, it can happen that it tries to insert the same key again. This results in a db error, which message is send as respone with http 200.

**To Reproduce**

Put load on service with parallel curl commands;

    for i in {1..5}; do curl -w'\n' -sSiv --raw "https://<host>/yourls-api.php?action=shorturl&format=simple&signature=<sig>&url=http%3A%2F%2F<url>" & done

You might want to increase the counter and/or try it several times.

**Expected behavior**

Catch the db exception, because it could be handled, either
 - retry or
 - respond with http status 500 (Internal server error) or maybe 503 (service unavailable)

Catch all unkown exceptions and respond with 500, log error.

**Actual behavior**

Response is the db error message with HTTP/1.1 200 OK

**Output**

    HTTP/1.1 200 OK

     <br />
    <b>Fatal error</b>:  Uncaught PDOException: SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry '1hwza' for key 'PRIMARY' in 
    /var/www/html/includes/vendor/aura/sql/src/ExtendedPdo.php:748
    Stack trace:
    #0 /var/www/html/includes/vendor/aura/sql/src/ExtendedPdo.php(748): PDOStatement-&gt;execute()
    #1 /var/www/html/includes/vendor/aura/sql/src/ExtendedPdo.php(306): Aura\Sql\ExtendedPdo-&gt;perform('INSERT INTO `yo...', Array)
    #2 /var/www/html/includes/functions.php(161): Aura\Sql\ExtendedPdo-&gt;fetchAffected('INSERT INTO `yo...', Array)
    #3 /var/www/html/includes/functions.php(278): yourls_insert_link_in_db('http://www.XXXX...', '1hwza', 'http://www.qweq...')
    #4 /var/www/html/includes/functions-api.php(21): yourls_add_new_link('http://www.XXXX...', '1hwza', 'http://www.XXXX...')
    #5 /var/www/html/includes/functions-plugins.php(150): yourls_api_action_shorturl(false)
    #6 /var/www/html/yourls-api.php(35): yourls_apply_filter('api_action_shor...', false)
    #7 {main}
      thrown in <b>/var/www/html/includes/vendor/aura/sql/src/ExtendedPdo.php</b> on line <b>748</b><br />

**Versions**
1.7.2 on Google Cloud Kubernetes

[...]

PR #3233, opened 2022-02-25
Try and catch concurrency errors

ozh:
When a keyword is supposed to be free and the INSERT throws a Fatal Error, return a 503 instead of a 200.

See #2538

Debatable : 
* we could consider returning a "503 Service Unavailable" because the server couldn't perform the operation due to concurrency
* or maybe a "409 Conflict" because the client requested something that is, after all, incorrect, even though it is unaware of it at the time of the request. For the record, 409 seems to be the standard response for concurrency issue.


ozh:
And damn that `yourls_add_new_link()` badly needs refactoring.

PR #3263, opened 2022-03-31
Update config-sample.php

techroy23:
added new words


dgw:
Thanks for your contribution. However, expanding the default list of "bad words" is not one of our goals. See previous discussion at #1991.

PR #3276, opened 2022-04-06
Added graph

3kh0:
I added a contributors graph with contrib.rocks


dgw:
"Contributers" is the wrong spelling.

That said, I'll let another maintainer decide to merge this. I'm not a fan of having too many external images embedded in readmes.

ozh:
Hello, thanks for this PR !

I'll second @dgw here, I don't think this adds value as the information is easily available on github itself, relies on an external service (and might even decrease the value of showing sponsors)

PR #3284, opened 2022-04-14
Regularly update GeoIP db

ozh:
Workflow to update geoip db every monday if a new version is available

### DIFF
--- a/.github/workflows/update-geoip.yml
+++ b/.github/workflows/update-geoip.yml
@@ -43,7 +43,7 @@ jobs:
           LOCAL_CTIME=$(stat -c %Y includes/geo/GeoLite2-Country.mmdb)
           echo "Remote: $REMOTE_CTIME ($(date -d @$REMOTE_CTIME))"
           echo "Local: $LOCAL_CTIME ($(date -d @$LOCAL_CTIME))"
-          [ $local_ctime -lt $remote_ctime ] && curl -s "$URL" | tar -zvx -C includes/geo/ --strip-components 1 --
+          [ $LOCAL_CTIME -lt $REMOTE_CTIME ] && curl -s "$URL" | tar -zvx -C includes/geo/ --strip-components 1 --
 
       - name: "Debug info: Show git status"
         run: git status -vv --untracked=all
--- a/.github/workflows/update-geoip.yml
+++ b/.github/workflows/update-geoip.yml
@@ -55,6 +55,8 @@ jobs:
 
       - name: Create pull request
         uses: peter-evans/create-pull-request@v4
+        env:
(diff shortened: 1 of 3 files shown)


.github/workflows/update-geoip.yml
```diff
@@ -0,0 +1,72 @@
...
+  workflow_dispatch:
+
+# Cancels all previous workflow runs for the same branch that have not yet completed.
+concurrency:
+  # The concurrency group contains the workflow name and the branch name.
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  certificate-check:
```
LeoColomb, line 25: It could be easier to maintain to have the two update jobs in a single workflow file

LeoColomb (changes requested):
Sorry meant to do that instead

.github/workflows/update-geoip.yml
ozh: Once everything works sure :)

.github/workflows/update-geoip.yml
```diff
@@ -0,0 +1,72 @@
...
+          body: |
+            Updated GeoIP database, last verified on ${{ steps.get-date.outputs.DATE }}.
+
+            Source: https://www.maxmind.com/en/account/login
+          labels: |
+            dependencies
+          reviewers: |
+            ozh
+            LeoColomb
+            dgw
```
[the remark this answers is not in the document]
ozh: I tried I realized I wrote it wrong :) see https://github.com/YOURLS/YOURLS/pull/3282/commits/7870c869357087c3e2c37c6844451d597405eb15#diff-ada20e574a1b16dfa05d0c19f046770d59b741fe25aeab224ef07982d9e329aaL78

ozh:
Hmm, there's something but we're not there yet : `git clone` changes local filetime to the time of checkout, so we'll always have fresher local files than the remote ones 😅 Will improve this part

.github/workflows/update-geoip.yml
ozh: On second thoughts I'm not so sure about this.
* The 2 workflows work differently and I think I'll have an easier time debugging when one breaks
* The cacert workflow is "safe" : before committing there's a checksum to make sure the file is valid. There is no such thing for the geoip workflow and in the possibility that the download may be corrupted we need to run unit tests on these PR

.github/workflows/update-geoip.yml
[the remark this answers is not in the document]
ozh: Using team-reviewer or reviewers doesn't work as expected alas

PR #3298, opened 2022-04-20
Update GeoIP with Github App

ozh:
Source: https://github.com/peter-evans/create-pull-request/blob/main/docs/concepts-guidelines.md#authenticating-with-github-app-generated-tokens

This should trigger workflows from the PR

### DIFF
--- a/.github/workflows/update-geoip.yml
+++ b/.github/workflows/update-geoip.yml
@@ -32,8 +32,8 @@ jobs:
         uses: tibdex/github-app-token@v1
         id: generate-token
         with:
-          app_id: ${{ secrets.TRIGGERTRIGGER_APP_ID }}
-          private_key: ${{ secrets.TRIGGERTRIGGER_PRIVATE_KEY }}
+          app_id: ${{ secrets.BOT_APP_ID }}
+          private_key: ${{ secrets.BOT_PRIVATE_KEY }}
 
       - name: Check if newer GeoIP DB
         env:
@@ -68,6 +68,5 @@ jobs:
             Updated GeoIP database, last verified on ${{ steps.get-date.outputs.DATE }}.
 
             Source: https://www.maxmind.com/en/account/login
-
           labels: |
             dependencies
--- a/.github/workflows/update-geoip.yml
+++ b/.github/workflows/update-geoip.yml
@@ -24,10 +24,12 @@ concurrency:
 jobs:
   certificate-check:
     name: "Check for updated GeoIP DB"
+
     # Don't run the cron job on forks.
-    if: ${{ github.event_name != 'schedule' || github.repository == 'YOURLS/YOURLS' }}
-    # Don't run on a branch created by the workflow.
-    if: ${{ github.ref != 'refs/heads/auto-update-geoip' }}
+    # Don't run on a branch that was created by the workflow
+    if: |
+      ${{ github.event_name == 'schedule' && github.repository != 'YOURLS/YOURLS' }} &&
+      ${{ github.ref != 'refs/heads/auto-update-geoip' }}
 
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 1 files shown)


LeoColomb:
@ozh This creates a lot of noise right now... Do you need help?

ozh:
@LeoColomb yes :)

Problem as you may have figured out already : 
- PR correctly created when remote binary is updated, in branch `auto-update-geoip`
- the workflows are now successfully triggered in the PR (no need to close/reopen the PR)
- but the geoip workflow triggers again, comparing  `auto-update-geoip` with  `auto-update-geoip` and thus, since there is no change, automatically closes the PR

LeoColomb:
> * PR correctly created when remote binary is updated, in branch `auto-update-geoip`

👍 

> * the workflows are now successfully triggered in the PR (no need to close/reopen the PR)

Really nice to use an app now!
The usual workaround is to use a [personal access token](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/creating-a-personal-access-token), but an app is even better.

> * but the geoip workflow triggers again

Do you really need to run the workflow when `includes/geo` is changed?

https://github.com/YOURLS/YOURLS/blob/16f31e58870452fc2c88cc3bec01f31e76339da8/.github/workflows/update-geoip.yml#L14

ozh:
> Do you really need to run the workflow when `includes/geo` is changed?

Hmmm, good question. What do we really need ? Cron to periodically check and manual to test, but maybe all other triggers are superfluous?

.github/workflows/update-geoip.yml
```diff
@@ -24,14 +16,25 @@ concurrency:
 jobs:
   certificate-check:
     name: "Check for updated GeoIP DB"
+
     # Don't run the cron job on forks.
-    if: ${{ github.event_name != 'schedule' || github.repository == 'YOURLS/YOURLS' }}
+    # Don't run on a branch that was created by the workflow
+    if: |
+      ${{ github.event_name == 'schedule' && github.repository != 'YOURLS/YOURLS' }} &&
+      ${{ github.ref != 'refs/heads/auto-update-geoip' }}
```
LeoColomb, line 24:
```suggestion
```

All this block can now be dropped with confidence

.github/workflows/update-geoip.yml
```diff
@@ -24,14 +16,25 @@ concurrency:
...
     steps:
       - name: Checkout code
         uses: actions/checkout@v3
 
+      - name: Generate token
+        uses: tibdex/github-app-token@v1
+        id: generate-token
+        with:
+          app_id: ${{ secrets.TRIGGERTRIGGER_APP_ID }}
+          private_key: ${{ secrets.TRIGGERTRIGGER_PRIVATE_KEY }}
```
LeoColomb, line 36: Are this secrets registered at repo level or at org level?
Can I rename the app to "YOURLS Bot" in order to use it on other repos as well?

.github/workflows/update-geoip.yml
```diff
@@ -5,14 +5,6 @@ on:
...
     - cron: '37 13 * * 1'
   # Run on every push to `master`
-  push:
-    branches: [ master ]
-  # And whenever this workflow is updated or a PR attempts to update the GeoIP files.
-  pull_request:
-    paths:
-      - '.github/workflows/update-geoip.yml'
-      - 'includes/geo/*'
-  # And manually
```
LeoColomb, line 15: 👍 
(maybe the inline comment should be updated as well, or simply dropped)

LeoColomb (changes requested):

.github/workflows/update-geoip.yml
ozh: I put it at repo level but sure it can be renamed and moved. Is this even possible, not sure

LeoColomb: Done 👍

.github/workflows/update-geoip.yml
LeoColomb: It will run, but it won't interact with upstream repo, and will actually fail because the secrets won't be available there.

ozh: Yep, got that while working on ozh/YOURLS to lessen the noise here ;-)

LeoColomb (approved):

PR #3339, opened 2022-05-16
List required/suggested PHP extensions in composer.json

dgw:
Best-effort attempt to declare all extensions used by YOURLS.

Inspired by discussion on YOURLS/docs#42. Based on the output of https://github.com/RogerGee/php-ext-depends

`composer check-platform-reqs` will show the status of all except those in "suggest", which is a limitation to sort out later.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,7 +17,10 @@
     },
     "require": {
         "php": ">=7.4",
+        "ext-filter": "*",
+        "ext-hash": "*",
         "ext-pdo": "*",
+        "ext-zlib": "*",
         "ozh/bookmarkletgen": "^1.2",
         "rmccue/requests" : "^2.0",
         "pomo/pomo" : "^1.4",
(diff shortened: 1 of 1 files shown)


ozh (commented):
[RogerGee/php-ext-depends](https://github.com/RogerGee/php-ext-depends) goes a bit too far and doesn't handle well the cases of "if function exists do this, else here is the replacement"

Strictly speaking I think we **need** only these : 
* php >= 7.4
* ext-dom
* ext-pcre
* ext-pdo
* ext-pdo_mysql
* ext-filter
* ext-hash

And we **recommend**, mostly for performance : 
* ext-curl
* ext-json
* ext-openssl
* ext-zlib
* ext-iconv
* ext-mbstring
* ext-bcmath
* ext-posix (really not sure about this one but it should not hurt to mention)

dgw:
@ozh double-checked most of those and amended the patch. Of note: bcmath _or_ GMP can be used, so I listed both as suggested.

ozh (approved):

Issue #2102, opened 2016-05-31
Used to work - can no longer add edit short urls - nothing changed


Kiroteryn:
I searched for other closed/open issues related, and they were closed without resolution: 
1. https://github.com/YOURLS/YOURLS/issues/1887
2. https://github.com/YOURLS/YOURLS/issues/2089
3. Hmm I thought I saw a 3rd one, but just about to run out the door, will edit later

Perhaps someone could offer some insight? I can confirm there was nothing in console, even with debug mode on (no CORS blocks or anything like that). 

I checked out the site to get things working or I wouldn't have been able to add/edit/view things to begin with :) I have about 3 pages of links. It started when I went to edit a link, then that's when I realized.. nothing works! Funny enough, you can still delete. **I can view and delete, but not edit or add.** Strange, right? 

**Issue 1887 (#1 link above) has the EXACT issue as me -- worked fine, suddenly doesn't work with same reproduced results.**

The last activity performed that I recall was adding "blog3" shorturl to a tumblr site. Doesn't seem too iregular : https://myblogtumblr.com/post/12345679/url-is-quite-long-with-dashes-like-this

I don't know if it was triggered by that or when I went to edit a random entry that's related to another social site - nothing special or no odd naming conventions or special chars.

There is one consistent thing I noticed: Every open issue, the same mod closed it blaming the hosting service: How about instead of closing this ticket, let's work together and find the solution, hm? This seems to be reproduced lately.

**Another thing to add that seems quite important:** I reinstalled using a new db, and seemed to work again. I curiously swapped back to the old DB to see if that would work. Everything was back AND I could add things **.. and a few days later, this issue reproduced identically: Can suddenly, randomly, no longer add/edit.**
### Technical details
- YOURLS version: Latest
- PHP version: 5.6
- Win10 with multiple stable browsers
### Reproducible Bug Summary
1. Install yourls and setup db
2. Run yourls successfully for a few weeks
3. Out of nowhere, I cannot change, add links. No changes at all. I almost wish I _did_ change something so I knew what caused this.

This is a bug because you can no longer add/edit urls  or you get a loading GIF that never goes away.

chtaube:
What server software are you running? Is it a Windows server running IIS?

ozh:
I assume that when you say "I cannot add or edit", you get an infinite loading gif. Check the AJAX error that's returned (using Chrome devtools or Firebug for Firefox)

dylanh724:
He said

[...]

PR #3351, opened 2022-05-25
Fix broken edit/delete buttons on some occasions.

josephalvis:
This issue is described in:
Used to work - can no longer add edit short urls - nothing changed #2102

At least one cause of this can be if a reverse proxy (e.g. cloudflared tunnel) sends an "id" parameter on every request, the nonce verification for the edit and delete buttons breaks.

This change is to rename the parameter to "elid" to prevent it conflicting with the "id" param. This resolves the issue and the buttons work again.

### DIFF
--- a/admin/admin-ajax.php
+++ b/admin/admin-ajax.php
@@ -23,19 +23,19 @@ switch( $action ) {
 		break;
 
 	case 'edit_display':
-		yourls_verify_nonce( 'edit-link_'.$_REQUEST['id'], $_REQUEST['nonce'], false, 'omg error' );
+		yourls_verify_nonce( 'edit-link_'.$_REQUEST['elid'], $_REQUEST['nonce'], false, 'omg error' );
 		$row = yourls_table_edit_row ( $_REQUEST['keyword'] );
 		echo json_encode( array('html' => $row) );
 		break;
 
 	case 'edit_save':
-		yourls_verify_nonce( 'edit-save_'.$_REQUEST['id'], $_REQUEST['nonce'], false, 'omg error' );
+		yourls_verify_nonce( 'edit-save_'.$_REQUEST['elid'], $_REQUEST['nonce'], false, 'omg error' );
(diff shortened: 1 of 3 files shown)


ozh:
Not sure I get it. Do you mean that each and every web application using `id` as a REQUEST parameter is broken by Cloudflared Tunnel because they overwrite the value ? If so, there has to be a configuration setting somewhere to avoid this and adapt the tunnel to the app instead of the other way round.

This is a slippery road I'm relunctant to get into : patch bits of YOURLS here and there, not to fix things, but to make sure it works with other tools.

josephalvis:
> Not sure I get it. Do you mean that each and every web application using `id` as a REQUEST parameter is broken by Cloudflared Tunnel because they overwrite the value ? If so, there has to be a configuration setting somewhere to avoid this and adapt the tunnel to the app instead of the other way round.

I identified the wrong culprit. This had nothing to do with cloudflared.

It looks like the issue I was having was caused by sharing a domain between yourls and a synology NAS. The Synology webui was accessed as nas.example.com and yourls as test.example.com for instance.  Synology sets a cookie "id" which gets shared and clashes with the id used by yourls because $_REQUEST['id'] includes $_COOKIE['id'].

What would you think about excluding cookies when looking up 'id'? I'm no longer using the same domain for yourls and the NAS but it seems like it could happen again to someone else.

ozh:
> The Synology webui was accessed as nas.example.com and yourls as test.example.com for instance

I think this is the really problem here. A cookie set on nas.example.com should not be readable from test.example.com and vice versa -- at least cookies set on yourls1.example.com and yourls2.example.com should not be readable from an other YOURLS instance than the one it was created by.

Back to you suggestion, yes we could exclude cookies here, I was using $_REQUEST to handle either $_GET or $_POST (and we need this) but didn't think $_COOKIES could ever be problematic. How do you suggest we do this ? $_REQUEST is used probably hundreds of times in YOURLS and I don't want to create more side effects than potential problem maybe solved.

PR #3365, opened 2022-06-08
Wrapped `$user` in `preg_quote` function

### LINKED ISSUE #3364: Auto encrypting password does not work if username contains a square bracket
### Code of Conduct

### Submission validity

### Self troubleshooting

### Version

1.9.1

### Description

If a user's name includes square brackets, the password will not be encrypted and a preg_replace error is reported.

### Expectation

The password to be encrypted without any issue.

### Reproduction steps

Add

```php
$yourls_user_passwords = array(
    'testuser[2]' => 'password',
);
```

### Context

Seems like the bracket might be breaking something for the `preg_replace` pattern.

https://github.com/YOURLS/YOURLS/blob/3b69f2be52fdfa08c54eccdcb7b3cfc55560a851/includes/functions-auth.php#L216

I only tried curly braces as an additional test, which did work, but even just 1 square bracket breaks the encryption.

SXN31:
This pull request is to fix #3364 where a username with square brackets will prevent encryption of the password. Thanks to ozh for showing me the way!


LeoColomb (changes requested):
Thanks!
Just a styling improvement 🔍

ozh:
Nice and fast, thanks :)

What would be really cool would be to update unit tests as well, in [tests/data/auth/config-test-auth.php](https://github.com/YOURLS/YOURLS/blob/master/tests/data/auth/config-test-auth.php), so we make sure that in the future this edge case is still correctly dealt with

SXN31:
> 

Just so I'm understanding the test enough, is it sufficient to add another line with a user with brackets and a random password into the array?

ozh:
I'm quite lagging behind several PR's, sorry for that. Will look into this asap

PR #3367, opened 2022-06-12
Create index.html

arpan20032009:
added this to redirect to the admin page instead of seeing the files


ozh:
From the readme : [There is no index page at the root of the install](http://yourls.org/#FAQ;noindex).

arpan20032009:
i added it because i used it 1 time and saw the i i open the root dir in browser it showed me all the file contents
so i added the index.html to redirect to the admin portal

Issue #3427, opened 2022-09-21
Optimise database migrations and allow them to be run from CLI


ashleykleynhans:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/YOURLS/.github/blob/main/CODE_OF_CONDUCT.md).

### Submission validity

- [X] This is not a personal support request, that should be posted on [discussions community](https://github.com/YOURLS/YOURLS/discussions).
- [X] I checked [current issues](https://github.com/YOURLS/YOURLS/issues) and this request isn't a duplicate of an existing issue, opened or closed.

### Summary

Combine ALTER statements on the same table into a single statement to improve efficiency on huge database tables, and provide a tool to run the updates from CLI instead of the web browser to prevent timeouts in the browser.

### Motivation

1. When you have a huge DB table (more than 11 million entries in my case), running multiple `ALTER` statements on the same table takes an hour for each statement to run.
2. The upgrade process takes too long when you have a huge DB, causing the browser to time out.

### Describe the solution

1. When you are performing multiple `ALTER` statements on the same table, these could be combined into a single `ALTER` statement and the upgrade process would be much more efficient.
2. Provide an option to use a CLI upgrade tool to complete the upgrade process rather than having to do it from within a web browser.

PR #3429, opened 2022-09-28
Support for DB upgrades from CLI and performance improvements

ashleykleynhans:
Fixes for #3427:

- Updated `yourls_upgrade_to_506()` to perform a single ALTER on the same table to improve efficiency.
- Added `yourls_upgrade_to_507()` to add additional indexes (admin page takes over a minute to count clicks on a large DB).
- Added support for upgrading DB from the CLI because a web browser times out when the DB is large.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@
 *.config
 *.conf
 [Rr]obots.txt
+.maintenance
 
 # Ignore everything in user/
 user/*
--- a/cli/upgrade.php
+++ b/cli/upgrade.php
@@ -1,0 +1,53 @@
+<?php
+define( 'YOURLS_ADMIN', true );
+define( 'YOURLS_UPGRADING', true );
+require_once( dirname( __DIR__ ).'/includes/load-yourls.php' );
+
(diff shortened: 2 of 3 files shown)


ozh:
Hello

That's a very interesting proposal

A couple remarks : 

This would be much better suited for a plugin or 3rd party tool. That is, modify core `functions-upgrade.php` as required but make the CLI part itself an external tool.
Several reasons : 
- 95% of people don't know how to use a CLI, don't need, or just don't have access to (basic share hosting for instance). The core philosophy of the project is to bundle things that everybody needs, and allow external tools to be plugged in -- see https://github.com/YOURLS/awesome-yourls
- The CLI tool could get its own development pace (fixes, future devs and why not new features) without requiring releasing a new YOURLS version.

The simplification of the SQL query in `yourls_upgrade_to_506()` is valid : several changes are not included any longer in your patch

Side note concerning core patches : function naming need to follow our [code convention](https://docs.yourls.org/development/coding-standards.html#mandatory-naming-conventions)

PR #3431, opened 2022-10-03
When adding URLs, new HTML rows should have unique HTML IDs.

### LINKED ISSUE #3400: Adding several URLs gives same element HTML id to all new URLs
### Discussed in https://github.com/orgs/YOURLS/discussions/3399

<div type='discussions-op-text'>

<sup>Originally posted by **composertech** August  4, 2022</sup>
Hi,
I observed the following behaviour. I added several new URLs through the interface. In the same session I then clicked the delete icon beside an existing link. Result: the wrong URL was deleted. After refreshing the page, the behaviour is correct. Can anyone else verify this? Thanks.</div>

Indeed: all new URL are given element id `-yid1`.

barryhughes:
Addresses a problem that arises when adding a new URL.

Currently, when a new row is prepended to the table, it is given the id `#id-yid1`, which is the same ID as the first row upon page load. If further URLs are added, the problem repeats itself and we will end up with several rows all using `#id-yid1`. This can cause problems if the user attempts to delete one of those rows (consequently, the wrong row may be deleted).

Closes #3400.

### Testing instructions

demo-yourls-unique-ids-on-insert

1. Begin with two or more existing entries. Confirm each row has a unique ID (ie, `#id-yid1`, `#id-yid2`, etc).
2. Add some additional URLs. Confirm these also have unique IDs (previously, they did not).
3. Attempt to delete one of the newly added URLs. The correct URL should successfully be removed.

### Notes

- This solution is not bulletproof; IDs are filterable and therefore are not guaranteed to be numeric in nature (for similar reasons, counting the number of rows to determine the next ID number as I've done here is also not bulletproof): happy to add more robust logic if desired.
- I was a little unsure about code style, particularly whitespace in-between parens. That is, `( $foo, $bar )` or `($foo, $bar)` ... apologies if I missed any guidance about that. Also happy to adjust!

### DIFF
--- a/admin/admin-ajax.php
+++ b/admin/admin-ajax.php
@@ -18,7 +18,7 @@ switch( $action ) {
 
 	case 'add':
 		yourls_verify_nonce( 'add_url', $_REQUEST['nonce'], false, 'omg error' );
-		$return = yourls_add_new_link( $_REQUEST['url'], $_REQUEST['keyword'] );
+		$return = yourls_add_new_link( $_REQUEST['url'], $_REQUEST['keyword'], '', $_REQUEST['rowid'] );
 		echo json_encode($return);
 		break;
 
--- a/includes/functions-formatting.php
+++ b/includes/functions-formatting.php
@@ -52,12 +52,16 @@ function yourls_string2int($string, $chars = null) {
  * Return a unique string to be used as a valid HTML id
  *
  * @since   1.8.3
- * @param  string $prefix   Optional prefix
- * @return string           The unique string
+ * @param  string $prefix      Optional prefix
+ * @param  int    $initial_val The initial counter value (defaults to one)
+ * @return string              The unique string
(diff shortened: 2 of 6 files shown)


ozh:
Man, now that's a very neat PR. Detailed reasoning explained, screenshot for evidence, and more unit tests. So neat :)

Tests are failing, a trivial string stuff to fix:
```
ID is built using the specified prefix and counter value.
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'foo-10'
+'foo10'
```

barryhughes:
😬 Bah, apologies! Corrected in https://github.com/YOURLS/YOURLS/pull/3431/commits/ee1cd800ecb0d240e3f74317a5cc3a076c0cca36.

ozh:
🍕 !

PR #3433, opened 2022-10-04
Typofixes

nexxai:


admin/index.php
```diff
@@ -227,7 +227,7 @@
 		$display_on_page = ( $offset + 1 );
 	}
 
-	// Determing Total Amount Of Pages
+	// Determining Total Amount Of Pages
```
dgw, line 230: I see where you came from with the existing change, but in context (and based on the other comments above) this is the intended wording:

```suggestion
	// Determine Total Amount Of Pages
```

nexxai: Agreed.  Updated to use your suggestion.

dgw (approved):
LGTM

nexxai:
My apologies, I deleted the repo by accident but the changes should still be merged.

LeoColomb (approved):

PR #3474, opened 2022-12-09
Support PHP 8.2

LeoColomb:
- [x] Tests passing
- [x] Free of deprecation warnings
  * POMO will be updated separatly

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,7 +13,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php: ['8.0', '8.1']
+        php: ['8.0', '8.1', '8.2']
         phpunit: ['latest']
         include:
         - php: '7.4'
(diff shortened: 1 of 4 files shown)


ozh:
I'm surprised there are only 2 occurrences of `${something}` ?

LeoColomb:
And yet only 2 😁
https://github.com/search?q=repo%3AYOURLS%2FYOURLS+%22%24%7B%22+language%3APHP&type=code&l=PHP

ozh:
What about this one ? Did you leave it out on purpose, or just missed it ? https://github.com/YOURLS/YOURLS/blob/5434b6aec3dda7a0f6069e0dd8219b0a0f90f30a/yourls-infos.php#L314

LeoColomb:
@ozh Uhh, this is not a variable to string operation! (but a string to variable name)
😊

Sazzad-Mannan:
Hi i am getting 500 error when click on stats button in the home page.My site was created 4/5years ago.Recently i updated it with new version.Everything is working fine but stats page showing 500 error.What can causing the problem? Thanks in advance.

PR #3478, opened 2022-12-17
Use sandbox for all included files

### LINKED ISSUE #3321: Use sandbox to include every user defined PHP files
Since YOURLS 1.9 we have a foolproof sandbox to safely include PHP files, used for plugins : https://github.com/YOURLS/YOURLS/blob/27375bc813f2ddc38030800e466babb72d7a5a66/includes/functions-plugins.php#L682-L696

Rename this function to make it more generic, update tests accordingly, and reuse it for : 

Pages:
https://github.com/YOURLS/YOURLS/blob/27375bc813f2ddc38030800e466babb72d7a5a66/includes/functions-html.php#L865

Custom DB error:
https://github.com/YOURLS/YOURLS/blob/27375bc813f2ddc38030800e466babb72d7a5a66/includes/Database/YDB.php#L161

Custom maintenance page:
https://github.com/YOURLS/YOURLS/blob/27375bc813f2ddc38030800e466babb72d7a5a66/includes/functions.php#L1052

Custom DB or cache layer: 
https://github.com/YOURLS/YOURLS/blob/27375bc813f2ddc38030800e466babb72d7a5a66/includes/Config/Init.php#L143-L165
(by the way, rename that "cache.php" to something understandable -- `must-use-plugin.php`? `unconditional-use-plugin.php`?)

jarebear6expepjozn6rakjq5iczi3irqwphcvb:
- replaced functions-plugins.php/yourls_activate_plugin_sandbox() with functions.php/yourls_activate_file_sandbox()
- updated calls to functions

Closes #3321


ozh:
Thanks ! This is a very nice start

The improvements from there would be : 
- moving `yourls_activate_plugin_sandbox()` to `functions-deprecated.php`, that's what we do when we rename or delete a function. This way, existing 3rd party code using it still functions (sometimes partially at least) and provides notice to update code
- adding unit tests for `yourls_activate_file_sandbox()` itself
- clean up code : all the "_if file exists then yourls_activate_file_sandbox it_" are redundant, since the function checks if the file exists
- actually making use of the return value, like we do when activating the plugin or using the uninstall file : if something goes wrong, the return is used to display an error. I don't think this check would be useful in all scenarios - for instance not when loading multiple files in `yourls_load_plugins()`, but definitely when loading individual files. See for instance `dead_or_error()`: if the custom `db_error.php` is existing but contains broken code, your patch would end up displaying no error message at all

e-ht:
Let me know if this covers what you were thinking. There are a few cases where the error would be odd if returned, as you mentioned. Let me know if this is way off base.  Thanks.

Issue #3484, opened 2022-12-27
Editing entry in list will write changes back to first entry


4BRobby:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/YOURLS/.github/blob/main/CODE_OF_CONDUCT.md).

### Submission validity

- [X] This is not a personal support request, that should be posted on [discussions community](https://github.com/YOURLS/YOURLS/discussions).
- [X] I checked [current issues](https://github.com/YOURLS/YOURLS/issues) and this request isn't a duplicate of an existing issue, opened or closed.

### Self troubleshooting

- [X] I've read the [documentation](https://yourls.org/) and made sure to follow it.
- [X] I've read the [troubleshooting first steps](https://docs.yourls.org/guide/troubleshooting/first-steps.html) and [frequent issues](https://docs.yourls.org/guide/troubleshooting/frequent-issues.html) guides.

### Version

1.9.1

### Description

When editing not the first url in list (e.g. second) and saving it, the changed entry will be placed on top of the list.

When refreshing the page it shows that it is saved correctly.

Javascript bug when saving.

### Expectation

When saved the correct line is being updated.

### Reproduction steps

1. Go to Admin
2. Have at least two entries
3. Edit second entry inline and hit save
4. Will be placed in the first line

### Context

Investigation:
- yourls_table_edit_row is handling the keyword only
- it calls yourls_unique_element_id without any id parameter thus always returning the id for the first row
- html of the ajax request for getting edit html will always returning the edit row with the first result

Suggested solution:
- call yourls_table_edit_row with optional parameter (id / row id)
- in yourls_table_edit_row call yourls_unique_element_id with parameter

ozh:
Closed via #3485

PR #3485, opened 2022-12-27
Issue #3484 Fix edit will write first entry only

4BRobby:
Fix for Issue #3484 
Id will be passed to yourls_table_edit_row resulting in the correct line in the table to be updated.
Also the nonce will be calculated correctly now for this line.


ozh:
Nice, thanks a lot.

PR #3492, opened 2023-01-04
Add backwards compatibility to yourls_salt function (#3491)

### LINKED ISSUE #3491: Secret signature token for Secure passwordless API calls changes when upgrading from 1.7.4 to 1.9.1
### Code of Conduct

### Submission validity

### Self troubleshooting

### Version

1.9.1

### Description

When using the secure passwordless API to create short links, the signature changes when upgrading from 1.7.4 to 1.9.1.

```
https://server-hostnameyourls-api.php?action=shorturl&format=json&url=https://example.com&title=link&signature=1234567890
````

So for example in 1.7.4, my signature was 1234567890, but now in 1.9.1, it is different, and now my RabbitMQ consumers are failing when trying to shorten links as a result.

### Expectation

When upgrading from one version of YOURS to another, the secret signature token to make secure passwordless API calls should remain the same as it was in the previous version.

### Reproduction steps

1. Install YOURS 1.7.4.
2. Make note of your secret signature token for making secure passwordless API calls.
3. Upgrade to YOURS 1.9.1.
4. Observe that the secret signature token has changed.

### Context

* I was running Ubuntu 18.04 LTS with YOURLS 1.7.4, and upgraded to Ubuntu 22.04 LTS for YOURLS 1.9.1.
* I was using PHP 7.2 with YOURLS 1.7.4, and upgraded to PHP 8.1 for YOURLS 1.9.1.

### PR DESCRIPTION
**Pull request by ashleykleynhans**
Provides backwards compatibility for the `yourls_salt` function, so that people who are already using `md5` can continue to do so by adding the following line of code to their config.php file:

```php
define ( 'YOURLS_SALT_FUNCTION', 'md5' );
```

Resolves #3491 .

### DIFF
--- a/includes/functions-auth.php
+++ b/includes/functions-auth.php
@@ -597,6 +597,11 @@ function yourls_tick() {
  */
 function yourls_salt( $string ) {
 	$salt = defined('YOURLS_COOKIEKEY') ? YOURLS_COOKIEKEY : md5(__FILE__) ;
+
+	if (defined('YOURLS_SALT_FUNCTION') and in_array(YOURLS_SALT_FUNCTION, hash_algos())) {
+		return yourls_apply_filter('yourls_salt', (YOURLS_SALT_FUNCTION)($string . $salt), $string);
+	}
+
 	return yourls_apply_filter( 'yourls_salt', hash_hmac( yourls_hmac_algo(), $string,  $salt), $string );
 }
 


ozh:
This isn't needed, the filter already provides this backward compat possibility, if needed, without the need of another constant. (Rule of thumb: we would rather remove constants than add new ones)

Simply make a simple custom plugin to pass 'md5' (or any algo) to the filter 'hmac_algo' (see https://github.com/YOURLS/YOURLS/blob/master/includes/functions-auth.php#L589-L616)

Something like the following should work : 

```php
/*
Plugin Name: md5 salt
Description: (ಠ_ಠ)
Version: 0.1
Author: Ozh
*/

yourls_add_filter('hmac_algo', 'letmeusemd5iloveit');

function letmeusemd5iloveit() {
    return 'md5';
}

// well I think that's it
```

ashleykleynhans:
Thanks for your suggestion but the plugin does not work because the new `hash_hmac` and  old `md5` function return totally different results.

For example:

```php
$salt = 'somethingrandom';
$username = 'username';
$hash1 = substr(hash_hmac('md5', $username, $salt), 0, 10);
$hash2 = substr(md5($username . $salt), 0, 10);

echo $hash1 . PHP_EOL;
echo $hash2 . PHP_EOL;
```

Outputs:

```
35989f3656
e189d57b25
```

And they need to be the same

ashleykleynhans:
This did the trick though:

```php
yourls_add_filter('yourls_salt', 'legacy_md5_salt');

function legacy_md5_salt($salted_hash, $original_valye) {
    $salt = defined('YOURLS_COOKIEKEY') ? YOURLS_COOKIEKEY : md5(__FILE__) ;
    return md5($original_valye. $salt);
}
```

PR #3507, opened 2023-02-04
Replace deprecated uses of `set-output` in Actions

dgw:
Tin. Deadline is May 31.

Announcement: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

### DIFF
--- a/.github/workflows/update-certificates.yml
+++ b/.github/workflows/update-certificates.yml
@@ -39,7 +39,7 @@ jobs:
 
       - name: "Get date"
         id: get-date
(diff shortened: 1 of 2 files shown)


dgw:
Seems like PHPUnit needs some TLC, too. Version 10 appears to break YOURLS' tests, and doesn't work at all on PHP 8.0.

LeoColomb (approved):

LeoColomb:
Indeed, php8.0 is deprecated, I won't be surprised PHPUnit 10 drops support for it.

PR #3508, opened 2023-02-04
Use the correct version 9 of PHPUnit

LeoColomb:
Fix current GitHub Actions builds


dgw (approved):

ozh:
Some inconsistencies on GH : it will randomly not find `phpunit`, in any PHP version : 
image

Re-running the job fixes the problem.

PR #3517, opened 2023-02-12
Make Googlebot indexing filterable

### LINKED ISSUE #3486: Make Googlebot indexing filterable
People are unhappy that Google indexes their shortlinks (#2202 and others). Then people are unhappy that Google does not index their shortlinks (#2970 and others).

Put the google indexing logic into its own function and make it filterable so everybody can use the plugin they want

https://github.com/YOURLS/YOURLS/blob/master/includes/functions.php#L284-L287

ozh:
Closes #3486

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -264,8 +264,7 @@ function yourls_redirect( $location, $code = 301 ) {
  * Redirect to an existing short URL
  *
  * Redirect client to an existing short URL (no check performed) and execute misc tasks: update
- * clicks for short URL, update logs, and send a nocache header to prevent bots indexing short
- * URLS (see #2202)
+ * clicks for short URL, update logs, and send an X-Robots-Tag header to control indexing of a page.
  *
  * @since  1.7.3
  * @param  string $url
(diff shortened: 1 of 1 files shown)


ozh:
As of 1.9.2 (soon released as of writing) YOURLS default behavior is : 
```php
header( "X-Robots-Tag: noindex", true );
```

If you are unhappy with this, the following plugin will make you happy : 

```php
/*
Plugin Name: Send "X-Robots-Tag: index" header on redirection
Description: just as the name says
Version: 0.1
Author: Ozh
*/

yourls_add_filter( 'robots_tag_header', 'ozh_send_robots_tag_header_index_instead_please' );

function ozh_send_robots_tag_header_index_instead_please() {
    return 'index';
}
```

ozh:
As of 1.9.2 YOURLS default behavior is :

header( "X-Robots-Tag: noindex", true );

Akila3:
so in other words, if I don't want google to index my links, I don't need to change anything? any settings? all out of the box configured that way by default?

jarebear6expepjozn6rakjq5iczi3irqwphcvb:
true

jamiers99:
> true

Is this true of the main YOURLS admin login page as well?

Issue #3523, opened 2023-02-15
Bookmarklet strips `+` in long URL


ozh:
### Discussed in https://github.com/orgs/YOURLS/discussions/3522

<div type='discussions-op-text'>

<sup>Originally posted by **LtCmdrKeene** February 14, 2023</sup>
I discovered an issue when I use the YOURLS browser bookmarklet tool to quickly create a shortlink to the page I'm viewing.

I have the bookmarklet "installed" to my browser toolbar with this code as expected (directly from the YOURLS tools page).

When I create a link with it, any + marks that existed in the original URL are removed.
So https://www.bing.com/images/search?q=list+all+modes+screenshot&form=HDRSC4&first=1
becomes https://www.bing.com/images/search?q=listallmodesscreenshot&form=HDRSC4&first=1
which is of course broken.

If I copy the original URL and create it in YOURLS myself by pasting in the admin page, the created link does not lose the spaces.

This makes me believe there is a flaw somewhere hidden in the bookmarklet code, but I'm not skilled enough to know how to debug that.</div>

PR #3527, opened 2023-02-23
fixes issue #3523 'Bookmarklet strips + in long URL'

jarebear6expepjozn6rakjq5iczi3irqwphcvb:
removed urldecode() from bookmarklet $_GET requests.  $_REQUEST superglobals come decoded per https://www.php.net/manual/en/function.urldecode.php#refsect1-function.urldecode-notes

When decoded multiple times it can cause character issues, eg. removing valid '+' characters.

fixes issue  Bookmarklet strips + in long URL #3523


dgw (approved):
Indeed, the documentation is pretty clear:

> **Warning** The superglobals `$_GET` and `$_REQUEST` are already decoded. Using `urldecode()` on an element in `$_GET` or `$_REQUEST` could have unexpected and dangerous results.

ozh:
Thanks !

PR #3589, opened 2023-07-12
Fix results with 0 clicks on search

JensSpanier:
There is currently a bug where the search result doesn't return entries whith 0 clicks.

When using the search on the admin page, `$_GET['click_limit']` is alwas set. It's just empty. So I check for `!empty` instead of `isset`.

It's also a possible fix to remove `=` from the check `intval($_GET['click_limit']) >= 0`.

Let me know what you think about this.


ozh:
Thanks !

Issue #3627, opened 2023-09-19
`login_failed` hook is misleading; it triggers for any unauthenticated page load


dgw:
_Originally raised in [discussion #3626](https://github.com/orgs/YOURLS/discussions/3626)_

The `login_failed` hook is [called within `yourls_is_valid_user()`](https://github.com/YOURLS/YOURLS/blob/0e490dffb4d13fbd54fdfc55a6ecc1ecfa33e01d/includes/functions-auth.php#L115-L116) any time `$valid` is False. This is the case if checking the username/password combination fails, or if the API request token/signature is invalid, but also if the user has not tried to authenticate yet (e.g. they are just loading the login page, or being redirected from an admin link because they have not signed in yet).

For plugins that want to do something _only_ when someone attempts to log in and fails, there is no other more appropriate hook, and they must figure out how to handle "failed logins" when there wasn't even a login attempt.

YOURLS already has `pre_login_<type>` hooks that fire before a particular kind of authentication check is run (API signature, username/password, cookie verification, etc.) but only the single `login_failed` hook to catch all cases. Maximum backward compatibility (to protect against [spacebar heating](https://xkcd.com/1172/) complaints) suggests the following approach:

- Keep `login_failed` as-is for the current major version (1.x)
- Add a new, differently-named hook that fires _only_ if some form of authentication attempt was present in the request _and_ the provided credential(s) failed verification—maybe `invalid_login`?
- Add more specific hooks that fire if/when each authentication type fails, e.g. `invalid_login_username_password`, `invalid_login_signature`, `invalid_login_cookie`, etc.)

@YOURLS/core I look forward to reading your opinions of this idea. 🙂

LeoColomb:
@XHyperDEVX One line, sure, but with great giant responsibilities! 😊
I don't think this change is viable, as there are many situations you'd like to shunt the process when the login is _about_ to fail.
I'd advocate for a new, differently-named hook `invalid_login`-ish. 👍

PR #3628, opened 2023-09-20
Fix login_failed hook problem

XHyperDEVX:
Fix for #3626 and #3627


dgw:
This is fine if the hook is _only_ supposed to run after a user/pass login attempt. I wouldn't want to change behavior like this in a patch version, though.

@ozh?

Issue #3659, opened 2023-11-03
Referral tracking parameter


jaredkotoff:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/YOURLS/.github/blob/main/CODE_OF_CONDUCT.md).

### Submission validity

- [X] This is not a personal support request, that should be posted on [discussions community](https://github.com/YOURLS/YOURLS/discussions).
- [X] I checked [current issues](https://github.com/YOURLS/YOURLS/issues) and this request isn't a duplicate of an existing issue, opened or closed.

### Summary

Track a parameter on the end of short urls to allow tracking of where those link were clicked from.

Examples: `myourls.com/hello?ref=qrcode` or `myourls.com/hello?ref=discord`
When these links are accessed admins can see how many time a link has been clicked based on the `ref` parameter. This is helpful to know where a shortened link was accessed.

### Motivation

Currently to do something similar you need to create duplicate short urls that all need to be unique. They also aren't dynamic. 
This is useful to see where traffic is coming from and where urls are possibly gaining traction.
Lastly, not every site or source will pass something of "referral" header so this is a great way to guarantee some method of tracking a link and its placement.

Site note: this is a feature on questionpro.com that i recently used and really liked being able to see where people were accessing a survey it put out giving me another piece of information into demographics.

### Describe the solution

Add a referral join-table
example:
<table>
<tr>
 <td>
 keyword
 <td>
ref
 <td>
count
<tr>
 <td>
hello
 <td>
discord
 <td>
4
<tr>
 <td>
hello
 <td>
qrcode
 <td>
3
</table>
And update UI to show break down based on ref param

LeoColomb:
Thanks for opening this issue @jaredkotoff.
Isn't this something manageable with a plugin?

jaredkotoff:
It could be, but it's also somewhat common feature of other url shortens.

And unless the database scheme has a place to track this sort of thing, I didn't see anything in the docs about plug-ins making changes to the database.  Wasn't sure if that was frowned upon.

jarebear6expepjozn6rakjq5iczi3irqwphcvb:
In my opinion a referring parameter would be a value-add to core. Personally it would provide me more meaningful data than having location stats.

ozh:
Sorry I'm a bit late in the discussion. What is @YOURLS/core thoughts about this ?

[...]

PR #3663, opened 2023-11-10
Add ref query parameter 

jaredkotoff:
https://github.com/YOURLS/YOURLS/issues/3659


includes/version.php
```diff
@@ -17,4 +17,4 @@
  * Must be a string of an integer.
  *
  */
-define( 'YOURLS_DB_VERSION', '506' );
+define( 'YOURLS_DB_VERSION', '507' );
```
jaredkotoff, line 20: Not sure how these db versions get assigned

PR #3669, opened 2023-11-24
Update readme.html

mohammadshg-ir:
Considering the change of the address of the blog page, why not update the link? And also the new link have SSL.


dgw (approved):
You're right, why not? 😁

PR #3679, opened 2023-12-03
Defer loading textdomain after plugins

ozh:
Currently the main textdomain is enforced before plugins load. This prevents user plugins from altering the global setting `YOURLS_LANG` as defined in the DB.

I *think* there would be no unwanted side effect if we loaded the language after plugins ?

### DIFF
--- a/includes/Config/Init.php
+++ b/includes/Config/Init.php
@@ -32,11 +32,6 @@ class Init {
             date_default_timezone_set( 'UTC' );
         }
 
-        // Load locale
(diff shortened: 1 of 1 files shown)


8MiYile:
This is actually very simple, if the webmaster has set the language in the configuration file, then the configuration file is the main, if not, try the language to let the plug-in load

8MiYile:
okay it's working 
so how time todo merge

ozh:
Old stuff but LGTM, thoughts @YOURLS/core ?

Issue #3682, opened 2023-12-12
Replace deprecated get_all_options filter with an action


RavanH:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/YOURLS/.github/blob/main/CODE_OF_CONDUCT.md).

### Submission validity

- [X] This is not a personal support request, that should be posted on [discussions community](https://github.com/YOURLS/YOURLS/discussions).
- [X] I checked [current issues](https://github.com/YOURLS/YOURLS/issues) and this request isn't a duplicate of an existing issue, opened or closed.

### Summary

Hi, this get_all_options filter has apparently been deprecated and does not allow getting all options anymore. And as far as I can figure, there is no alternative wrapper function available...

### Motivation

For effective persistant options caching, plugins like APC Cache and YAPCache depended on the get_all_options filter. This filter has apparently been deprecated and does not allow getting all options anymore. This is needed to populate the cache.

### Describe the solution

On includes/Database/Options.php line 78 there is currently `yourls_apply_filter('get_all_options', 'deprecated');`

What if that was replaced with `yourls_do_action('get_all_options', $options);` then a cache plugin could hook there and fetch all options at that time to populate the persistant cache.

RavanH:
Unless the idea is top move all actions (and deprecated filter) in the Options class to their corresponding wrapper functions? In that case, I'll be happy to work on that :)

PR #3683, opened 2023-12-12
Replace deprecated get_all_options filter with an action

RavanH:
Feature request #3682

PR #3687, opened 2023-12-13
Move all actions and filters from Options class to wrapper functions

RavanH:
Alternative solution to #3682


ozh:
@RavanH thanks for your submission - I'm unable at the moment to take time to look into these, but I will :)

ozh:
Issue fixed with #3683 -- thanks again for your submissions :)

PR #3747, opened 2024-04-15
Update plugins.php

Shyzer:
Updated old 404 link on line 164

### DIFF
--- a/admin/plugins.php
+++ b/admin/plugins.php
@@ -161,7 +161,7 @@ yourls_html_menu();
 
 	<h3><?php yourls_e( 'More plugins' ); ?></h3>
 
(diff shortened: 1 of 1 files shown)


dgw:
We are probably better off making the older link work as it used to. Changing this string would require language packs to be updated, as I understand the L10N system.

@LeoColomb How hard would it be to redirect `/awesome` on the new website to the Awesome List?

LeoColomb:
@dgw As hard as understanding why the YOURLS instance in backend is down 😅

dgw:
@LeoColomb Oh no! Well I guess that's why it broke 😂

dgw:
The backend issue has been fixed! 🥳

Thanks @Shyzer for helping us notice the problem.

PR #3756, opened 2024-05-07
Ensure all `statusCode`/`errorCode` API values are strings

### LINKED ISSUE #3755: Inconsistent data type of statusCode in API response
### Code of Conduct

### Submission validity

### Self troubleshooting

### Version

1.9.2

### Description

In the API response for shortening a URL, if the request is successful, the statusCode in the JSON response is an integer. But in the case of an error, like the URL already having been shortened, the statusCode is returned as a string in the JSON. 

### Expectation

I would think it should be consistently one data type or the other.

### Reproduction steps

Make a shorten API call for a new URL. The statusCode will be an integer in the response JSON. Now try to shorten the same URL again. In the error response the statusCode is now a string.

### Context

I'd expect that both should be either an integer OR a string.

dgw:
PHPDoc comments about the API response format say that both `statusCode` and `errorCode` are expected to be strings, but a lot of cases were returning ints instead. This could cause problems for JSON clients.

This should resolve #3755.

### DIFF
--- a/admin/index.php
+++ b/admin/index.php
@@ -157,7 +157,7 @@ if ( isset( $_GET['u'] ) or isset( $_GET['up'] ) ) {
 
 				// Deal with the case when redirection failed:
 				$return['status']    = 'error';
-				$return['errorCode'] = 400;
+				$return['errorCode'] = '400';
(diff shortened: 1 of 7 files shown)


LeoColomb (approved):
Sounds reasonable 👍

PR #3768, opened 2024-06-01
Update config-sample.php

dzg:
Added comment regarding `$yourls_reserved_URL` being required.


ozh:
Let's fix this with improving the code via the issue you opened, instead ?

PR #3827, opened 2024-11-26
Run CI on latest PHP versions including 8.4

LeoColomb:


LeoColomb:
https://github.com/auraphp/Aura.Sql/pull/231

Issue #2799, opened 2020-12-07
Fatal error with MySQL 8 due to double quotes ("")


paulgoodchild:
* [x] I've read the [Troubleshooting First Steps](https://github.com/YOURLS/YOURLS/wiki/Troubleshooting-first-steps)
* [x] This request isn't a duplicate of an existing issue, opened or closed
* [x] I've read the [docs](https://yourls.org/) and followed them (if applicable)
* [x] This is not a personal support request that should be posted on the [YOURLS Discourse](https://discourse.yourls.org/) community

**Describe the bug**
When using MySQL 8 (MySQL 5.x is unaffected) YOURLS encounters a fatal error when running a YOURLS search in the admin.

The error is:

```
[07-Dec-2020 15:59:29 UTC] PHP Fatal error:  Uncaught PDOException: SQLSTATE[42S22]: Column not found: 1054 Unknown column '' in 'where clause' in /includes/vendor/aura/sql/src/ExtendedPdo.php:752
Stack trace:
#0 /includes/vendor/aura/sql/src/ExtendedPdo.php(752): PDOStatement->execute()
#1 /includes/vendor/aura/sql/src/ExtendedPdo.php(485): Aura\Sql\ExtendedPdo->perform('SELECT COUNT(ke...', Array)
#2 /includes/functions.php(691): Aura\Sql\ExtendedPdo->fetchObject('SELECT COUNT(ke...', Array)
#3 /admin/index.php(153): yourls_get_db_stats(Array)
#4 {main}
  thrown in /includes/vendor/aura/sql/src/ExtendedPdo.php on line 752
```

**To Reproduce**
1. Install on MySQL 8
2. Search Yourls using 'search for' term in admin search function, no other parameters.

**Expected behavior**
Should return a results page for search

**Actual behavior**
Fatal error produced as above

**Versions**
Latest version 1.7.9

**Additional context**
The problem is alluded to here:
https://stackoverflow.com/questions/11321491/when-to-use-single-quotes-double-quotes-and-backticks-in-mysql

MySQL prefers single quotes for string literals, but does support double quotes in some circumstances. It appears that MySQL 8 is less forgiving when using double quotes for strings.

It turns out that the fix for this is trivial.
`admin/index.php`
Line: 64
Replace double quotes in `CONCAT_WS`:
```php
        $where['sql'] .= ' AND CONCAT_WS("",`keyword`,`url`,`title`,`ip`) LIKE (:search)';
```
with single quotes:
[...]

PR #3828, opened 2024-11-29
Mysql 8+ has issue with double quotes in install script

halkeye:
Same issue as https://github.com/YOURLS/YOURLS/issues/2799 but install file
Ran into the issue on mysql 8.0.30 on digitalocean hosted mysql solution.


LeoColomb (approved):
LGTM

PR #3830, opened 2024-12-02
Fix 0 access entry disappear when count threshold field is empty

infinitail:


includes/Views/AdminParams.php
```diff
@@ -222,7 +222,7 @@ public function get_click_filter()
     public function get_click_limit()
     {
         // @hook Default link click threshold (unset)
-        return (!empty($_GET['click_limit']) && intval($_GET['click_limit']) >= 0) ?
+        return (!empty($_GET['click_limit']) && $_GET['click_limit'] !== '' && intval($_GET['click_limit']) >= 0) ?
```
LeoColomb, line 225: As per PHP documentation, `empty($var)` is [equivalent](https://www.php.net/manual/en/function.empty.php) to `!isset($var) || $var == false`.
And [`(bool) $var` is `false` if `$var = ''`](https://www.php.net/manual/en/language.types.boolean.php#language.types.boolean.casting).

Except if I'm missing something, this addition does not alter the behavior in any way.

infinitail: When the form is submitted with empty value, get_click_limit() function returns '0'. However, the subsequent process in "admin/index.php" expects "$click_limit" to be empty string if not specified.

```admin/index.php
// Click filtering
$click_limit = $view_params->get_click_limit();
if ( $click_limit !== '' ) {
    $click_filter   = $view_params->get_click_filter();
```

LeoColomb: Thanks for your input.
While I understand your description, I insist on the fact this addition is not changing the behavior of the function at all.
As soon as the value is not `empty`, it will for sure not be `''`.

Please see https://3v4l.org/nPfZf for concrete running code.

infinitail:
@LeoColomb 

I'm very sorry, I found this issue in YOURLS 1.9.2 and created a patch but I submitted the pull request based on the master branch.
https://github.com/YOURLS/YOURLS/blob/1.9.2/includes/Views/AdminParams.php#L225

LeoColomb:
Oh indeed, this has been changed recently: #3589.
My mistake too, not to remember it.
It should be fixed by the next release, I'm closing your PR.

PR #3835, opened 2024-12-05
Remove remaining leading slash from request

Weax:
I am routing yourls to a url with subfoler using reverse proxy to example.com/to so links looks like example.com/to/request

On my server:
$_SERVER[ 'REQUEST_URI' ] returns "/request"
$yourls_site equals "/to/"

yourls_get_request function returns "/request" instead of "request".

$request starts with an initial slash. 
Looks like it safe just to remove any available slash at the beginning.

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -1025,6 +1025,7 @@ function yourls_get_request($yourls_site = '', $uri = '') {
         $request = current( explode( '?', $request ) );
     }
 
+    $request = ltrim( $request, '/' );
     $request = yourls_sanitize_url( $request );
 
     return (string)yourls_apply_filter( 'get_request', $request );


ozh:
YOURLS_SITE should **not** have a trailing slash, see https://github.com/YOURLS/YOURLS/blob/master/user/config-sample.php#L37

Weax:
it doesn't have

ozh:
I'm not sure I get it. Do we end up having something with double slash somewhere ?

ozh:
There is a failing test, but maybe the test is currently written to work with a flawed logic -- I'm not excluding that possibility :)

Closing because no feedback - feel free to comment if we should re open

PR #3838, opened 2024-12-14
Update CI badge syntax

janbrasna:
The badge would hint that CI is failing, when it's actually not. It's just the URL/API that needs updating, to the current form referring workflow files instead of workflow names for disambiguation..


LeoColomb (approved):
Thanks!

PR #3852, opened 2025-02-09
Upgrade Aura.SQL to fix PHP 8.4 compatibility

### LINKED ISSUE #3839: PHP 8.4 not supported
### Self troubleshooting

- [x] I've read the [documentation](https://yourls.org/) and made sure to follow it.
- [x] I've read the [troubleshooting first steps](https://docs.yourls.org/guide/troubleshooting/first-steps.html) and [frequent issues](https://docs.yourls.org/guide/troubleshooting/frequent-issues.html) guides.

### Version

1.9.2

### Description

I've updated php from 8.3 to 8.4 on a shared hosting and YOURLS stopped working throwing an Error 500. Sorry that I can't provide any logs because there is nothing found on logs about YOURLS error.

### Expected Behavior

To be PHP 8.4 compatible.

### Steps to Reproduce

1. Update PHP to 8.4 version
2. Try to access admin page or any shortened link
3. An Error 500 appears

### Context

_No response_

AlexanderWillner:
This pull request addresses issue https://github.com/YOURLS/YOURLS/issues/3839

Fixes #3839
Closes #3827


composer.json
```diff
@@ -27,7 +27,7 @@
         "rmccue/requests" : "^2.0",
         "pomo/pomo" : "^1.4",
         "geoip2/geoip2" : "^2.10",
-        "aura/sql": "^3.0",
+        "aura/sql": "^6.0.0",
```
LeoColomb, line 30: Let's keep only two digits for the version

composer.json
```diff
@@ -37,10 +37,7 @@
         "ext-ctype": "*"
     },
     "config": {
-        "vendor-dir": "includes/vendor",
-        "platform": {
-            "php": "7.4.0"
```
LeoColomb, line 42: Please keep the platform requirement. 
PHP must be the lowest version YOURLS and its dependencies support.

LeoColomb (changes requested):
Thanks for submitting this pull request, @AlexanderWillner. 
Looks great overall, see my online comments for some tweaking.

composer.json
AlexanderWillner: The Caret Version Range actually says: `>=6.0.0 <7.0.0`

composer.json
AlexanderWillner: Thanks - the lowest version YOURLS supports is automatically determined by the dependencies. In this case it would be PHP 8.x : https://github.com/auraphp/Aura.Sql/blob/6.x/composer.json#L25
I'd rather update https://github.com/YOURLS/YOURLS/blob/master/composer.json#L19 to specify that PHP 8.x is needed. Does this make sense?

composer.json
LeoColomb: Indeed, but streamlining to two digits somewhat eases the maintenance.

```suggestion
        "aura/sql": "^6.0",
```

composer.json
LeoColomb: That's correct, but the goal of a platform specification is the other way around: ensure the dependencies are _not_ updated to a version that breaks minimum requirements.
That super useful for automatic dependency management.

AlexanderWillner: I see. Makes sense. I pushed an update. Note that I've not pushed an update for `includes/vendor/`

LeoColomb (changes requested):
Thanks for your last edits @AlexanderWillner.
While your changes look good to me, I'm a bit triggered by the fact all previous PHP versions are simply dropped.
It feels like a giant step, somewhat too broad.
What do you think, @YOURLS/core?

dgw:
> What do you think, @YOURLS/core?

NAK, the software needs to support _at minimum_ all [currently-maintained PHP versions](https://www.php.net/supported-versions.php) (right now, 8.1-8.4). Dropping everything older than 8.4 is a no-go, and CI coverage for the whole range also needs to be kept.

I'm fine with formally dropping 7.4 (last updated 26 months ago) and _mostly_ fine with dropping 8.0 (last updated 14 months ago). My hosting company forcibly updated anyone still using 7.4, at least, some time ago. Any host that still has those dead versions around is encouraging bad practices, IMO. [[PHP EOL](https://www.php.net/eol.php)]

AlexanderWillner:
I agree, it should work with PHP 8.0 / PHP 8.1 - currently https://github.com/auraphp/Aura.Sql/blob/8e2bb362e8953198df3682c9122e8b9edab5ff20/composer.json#L25 points to >= PHP 8.4 - not sure if this is really required

AlexanderWillner (commented):
thanks, missed that

includes/vendor/composer/installed.json
```diff
@@ -2,28 +2,29 @@
     "packages": [
         {
             "name": "aura/sql",
-            "version": "3.1.0",
-            "version_normalized": "3.1.0.0",
+            "version": "6.0.0",
```
LeoColomb, line 5: I might have missed something, but are we using version 5 or version 6?

AlexanderWillner: The idea is to use version 5 as suggested by @srjlewis in https://github.com/auraphp/Aura.Sql/commit/92df92269e1961304d7c46ef0046d71b6f0f3fec#commitcomment-152736781. Version 5 supports PHP < 8.4

LeoColomb: Yes, that is clear 😊
But aren't the files in here coming from version 6?

AlexanderWillner: Yes, how would you like handle the changes in `includes/vendor/`? There would be around 80 changes such as:
- updating aura/sql
- updating modules due to the change in the PHP version
- removing old dependencies such as includes/vendor/symfony/polyfill-php72

Usually I would not include `includes/vendor` but use the composer files instead to generate this folder automatically.

srjlewis:
Hi all

To chip in, I do run a fork of aura/sql that is a direct replacement, but is compatable with < 8.4 and replaces v5 and v6 of aura/sql which I use withing of projects.

[srjlewis/aura-sql](https://packagist.org/packages/srjlewis/aura-sql)

I keep it as close to the origanal as posable but with the following changes https://github.com/auraphp/Aura.Sql/compare/6.x...srjlewis:Aura.Sql:6.x

Please fill feel free to use it, and I can make any changes with in reason.

LeoColomb:
Thanks for the input @srjlewis.
Seeing the diffs are pretty little, is there a specific reason for them not to be included in main Aura.SQL?

includes/vendor/composer/installed.json
LeoColomb: > Usually I would not include `includes/vendor`

Yes, I'd prefer too. But for history reasons, `includes/vendor` has to be part of the repo.
I'll do my best to find a correct resolution 😊

LeoColomb (approved):

srjlewis:
> Thanks for the input @srjlewis. Seeing the diffs are pretty little, is there a specific reason for them not to be included in main Aura.SQL?

There was some large discussions about it, but they decided they only want to support PHP 8.4 and above, and not use my sugestions.

PR #3863, opened 2025-03-03
Upgrade dependencies to match PHP 8.1+ requirement

LeoColomb:
Plenty of files listed in this PR, but the actual changes are pretty limited.
Direct follow up of #3852 

* Upgrade all dependencies based on PHP 8.1+ req
* Upgrade tests to support PHPUnit 10+
  * Dropping prior version of PHPUnit as per PHP < 8.0 support drop
  * Compatibility changes applied semi-automatically with Rector
  * Minor edits to ease test usage


readme.html
```diff
@@ -531,7 +531,7 @@ <h1><a href="https://yourls.org/">YOURLS: <span>Y</span>our <span>O</span>wn <sp
 
 			<a class="anch" href="#FAQ;requirement"><span>&#10148;</span><h3 id="requirement">Server requirements and recommendations</h3></a>
 			<ol>
-				<li>We recommend <strong>PHP 7.4</strong> or above</li>
+				<li>We recommend <strong>PHP 8.1</strong> or above</li>
```
dgw, line 534: Is it really still a recommendation, or is it a requirement now?

LeoColomb: I guess it's required now 😅
```suggestion
				<li><strong>PHP 8.1</strong> or above is required</li>
```

dgw:
I'm obviously not going to dive in and review all the vendor changes, test framework upgrades, etc. 😅

But I clearly need to learn more about Composer and `vendor/` handling, because aura.sql ^6.0 shouldn't work on PHP < 8.4 according to their own requirements.

ozh:
> because aura.sql ^6.0 shouldn't work on PHP < 8.4 according to their own requirements.

For real. I want to update spatie/array-to-xml but composer halts because 

```
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - aura/sql is locked to version 6.0.0 and an update of this package was not requested.
    - aura/sql 6.0.0 requires php ^8.4 -> your php version (8.1.0; overridden via config.platform, actual: 8.4.5) does not satisfy that requirement.
```

PR #3887, opened 2025-04-03
fix composer install fails for new version 1.10.0

faborg:
The PR fixes the following problem when running `composer install` on a fresh YOURLS-installation.

```
YOURLS git:(master) composer install
Installing dependencies from lock file (including require-dev)
Verifying lock file contents can be installed on current platform.
Your lock file does not contain a compatible set of packages. Please run composer update.

  Problem 1
    - aura/sql is locked to version 6.0.0 and an update of this package was not requested.
    - aura/sql 6.0.0 requires php ^8.4 -> your php version (8.1.0; overridden via config.platform, actual: 8.4.1) does not satisfy that requirement.
```


LeoColomb:
Thanks for submitting this pull request, @faborg.
The conflict between version is a bit more tricky to be honest and was a concern when merging Aura.SQL v6.
Out of curiosity, what brings you running a `composer install` on a YOURLS instance?

faborg:
The problem also occurs when running the create-project command like this: `composer create-project yourls/yourls=1.10.0 .` or just `composer create-project yourls/yourls .`

LeoColomb:
Could you try to run the following instead:

```shell
composer create-project --ignore-platform-req=php yourls/yourls .
```

Dherlou:
> Could you try to run the following instead:
> 
> ```shell
> composer create-project --ignore-platform-req=php yourls/yourls .
> ```

This seems to have solved it in my PHP 8.1 environment.

faborg:
Yes, with the `--ignore-platform-req=php` option it also works for me with PHP 8.4.

I'll keep this PR open for the moment, because I think it should work without the option.

Thanks for the help!

LeoColomb:
Thanks for the confirmation @faborg @Dherlou.

> because I think it should work without the option.

I fully agree in theory, but one of our dependencies raises version conflict that forces us to find a more pragmatic (and far from perfection) solution 😬

Dherlou:
@LeoColomb Is it possible that this problem also affects the v1.10/latest docker images? I had to downgrade another instance to v1.9.2-apache as the latest tag resulted in empty HTML output, although the container is running and receiving the requests.

PR #3871, opened 2025-03-07
Fix all PHPUnit tests errors and compatiblity issues

LeoColomb:
Quite ugly commit log, and not much prettier changes, but at least everything runs smoothly now.

PR #3893, opened 2025-04-06
Revert added exception in the sandbox

### LINKED ISSUE #3892: Unable to deactivate [default] plugins
### Self troubleshooting

- [x] I've read the [documentation](https://yourls.org/) and made sure to follow it.
- [x] I've read the [troubleshooting first steps](https://docs.yourls.org/guide/troubleshooting/first-steps.html) and [frequent issues](https://docs.yourls.org/guide/troubleshooting/frequent-issues.html) guides.

### Version

1.10.0

### Description

Once you've enabled one of the starter plugins, you cannot disable them. An error message appears on screen stating the `uninstall.php` does not exist.

### Expected Behavior

_No response_

### Steps to Reproduce

1. Activate a plugin on your instance.
2. Attempt to deactivate.
3. Receive error prompt.

### Context

https://github.com/mother-of-all-self-hosting/ansible-role-yourls/issues/4#issuecomment-2781276315

Image

LeoColomb:
Fixes #3892
Ref #3871

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -1316,8 +1316,6 @@ function yourls_include_file_sandbox($file) {
             require_once $file;
             yourls_debug_log("loaded $file");
             return true;
-        } else {
-            throw new \Exception('File not readable');
         }
     } catch ( \Throwable $e ) {
         yourls_debug_log("could not load $file");
--- a/tests/tests/utilities/FileLoaderTest.php
+++ b/tests/tests/utilities/FileLoaderTest.php
@@ -25,7 +25,7 @@ class FileLoaderTest extends PHPUnit\Framework\TestCase {
      * Load missing file = string
      */
     public function test_load_file_not_exists() {
-        $this->assertIsString( yourls_include_file_sandbox( YOURLS_TESTDATA_DIR . "/" . rand_str() . ".php" ) );
+        $this->assertNull( yourls_include_file_sandbox( YOURLS_TESTDATA_DIR . "/" . rand_str() . ".php" ) );
     }
 
     /**


dgw (approved):
Just fine on its face; #3871 changed `yourls_include_file_sandbox()`'s behavior when it probably didn't need to.

<details>
<summary>Deeper musings unrelated to making a hotfix</summary>

Hmm, from this comment in the function doc:

```
 * This function does not check first if the file exists : depending on use case, you may check first.
```

I would think that upstream code is expected to test for the file's existence before trying to `include_file_sandbox` it, here in between setting `$uninst_file` and `$attempt`:

https://github.com/YOURLS/YOURLS/blob/b3848d4833b15744af82a9b235b267cb90ad6186/includes/functions-plugins.php#L691-L703

Something like this, maybe (GH doesn't support creating suggestions outside the patched area of a PR, sadly)?

```php
    // Check if we have an uninstall file - load if so 
    $uninst_file = YOURLS_PLUGINDIR . '/' . dirname($plugin) . '/uninstall.php';
    $attempt = yourls_include_file_sandbox( $uninst_file );

    // Try to set plugin uninstall mode if valid
    if ( file_exists( $uninst_file ) {
        if ( $attempt === true ) {
            define('YOURLS_UNINSTALL_PLUGIN', true);
        }
    }

    // Check if we have an error to display
    elseif ( is_string( $attempt ) ) {
        $message = yourls_s( 'Loading %s generated unexpected output. Error was: <br/><pre>%s</pre>', $uninst_file, $attempt );
        return( $message );
    }
```

</details>

Issue #3695, opened 2023-12-29
Traffic statistics is incorrect in case of no recent activity


xhruso00:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/YOURLS/.github/blob/main/CODE_OF_CONDUCT.md).

### Submission validity

- [X] This is not a personal support request, that should be posted on [discussions community](https://github.com/YOURLS/YOURLS/discussions).
- [X] I checked [current issues](https://github.com/YOURLS/YOURLS/issues) and this request isn't a duplicate of an existing issue, opened or closed.

### Self troubleshooting

- [X] I've read the [documentation](https://yourls.org/) and made sure to follow it.
- [X] I've read the [troubleshooting first steps](https://docs.yourls.org/guide/troubleshooting/first-steps.html) and [frequent issues](https://docs.yourls.org/guide/troubleshooting/frequent-issues.html) guides.

### Version

1.9.2

### Description

Links are showing past days activity that is counted from last redirection date instead of counting it from current date. 

### Expectation

Last 7 days should show activity in last 7 days. Currently it shows last 7 days from last activity day.

### Context

Using PHP 8.2. I have attached picture where one can see that currently it is 29DEC23. Last activity date is 21DEC23 (from database). The included graph shows current date as 21DEC. 

I assume "yourls_build_list_of_days" should fill empty dates with 0 up to current date, not just in between dates

<img width="1125" alt="Screenshot 2023-12-29 at 18 56 14" src="https://github.com/YOURLS/YOURLS/assets/1449191/edf91259-8861-4545-a3c5-302ae9544e5d">

```

referrers: Array
(
)
referrer sort: Array
(
)
direct: 5
notdirect: 0
dates: Array
(
    [2023] => Array
        (
            [12] => Array
                (
                    [17] => 4
                    [21] => 1
                )

        )

)
list of days: Array
(
    [Dec 17, 2023] => 4
    [Dec 18, 2023] => 0
    [Dec 19, 2023] => 0
    [Dec 20, 2023] => 0
    [Dec 21, 2023] => 1
)
list_of_months: Array
(
    [12] => 12
)
list_of_years: Array
(
    [2023] => 2023
)
last_24h: Array
(
    [13 PM] => 0
    [14 PM] => 0
    [15 PM] => 0
    [16 PM] => 0
    [17 PM] => 0
    [18 PM] => 0
    [19 PM] => 0
    [20 PM] => 0
[...]

PR #3895, opened 2025-04-11
Extend stats timeline to current day

ntindicator:
Fixes issue  #3695 - Traffic statistics is incorrect in case of no recent activity. The stats timeline stopped at the last click. This change ensures users see stats up to today. Now pads $list_of_days with zeros from the last click to today in
yourls_build_list_of_days().
Tested with multiple URLs to confirm all timeframes display correctly.

### DIFF
--- a/includes/functions-infos.php
+++ b/includes/functions-infos.php
@@ -136,20 +136,27 @@ function yourls_build_list_of_days($dates) {
 
 	unset( $_keys );
 
+	// Extend to today
+	$today = new DateTime();
+	$today->setTime(0, 0, 0); // Start of today
+	$today_year = $today->format('Y');
+	$today_month = $today->format('m');
+	$today_day = $today->format('d');
+
 	// Now build a list of all years (2009), month (08 & 09) and days (all from 2009-08-29 to 2009-09-05)
 	$list_of_years  = array();
 	$list_of_months = array();
 	$list_of_days   = array();
-	for ( $year = $first_year; $year <= $last_year; $year++ ) {
+	for ($year = $first_year; $year <= $today_year; $year++) {
 		$_year = sprintf( '%04d', $year );
 		$list_of_years[ $_year ] = $_year;
 		$current_first_month = ( $year == $first_year ? $first_month : '01' );
-		$current_last_month  = ( $year == $last_year ? $last_month : '12' );
+		$current_last_month = ($year == $today_year ? $today_month : '12');
 		for ( $month = $current_first_month; $month <= $current_last_month; $month++ ) {
 			$_month = sprintf( '%02d', $month );
 			$list_of_months[ $_month ] = $_month;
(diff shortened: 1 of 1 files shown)


ntindicator:
The white space removal was accidental, only noticed it after I'd already pushed the commit. Fixed it now :)

LeoColomb (approved):
LGTM (but not tested).
@dgw Agreed 🚀

dgw:
> (but not tested)

Thankfully it's easy enough to copy-paste `functions-infos.php` over to my server to test it out.

I don't have a super active instance, so the highest-click-count example I can test ain't much. It's enough to confirm that this patch does what it says, though:

image

Compare to the old line chart, before I updated `functions-infos.php` from this PR branch:

image

Note that the "last 7 days" and "last 30 days" options disappeared after this patch, because there is no data within those time periods. That's _probably_ appropriate. Why invite the user to view empty data?

What worries me a little more is that the click total I get by summing the points in the updated line graph view is only about half the actual total reported in the table to the right. (Probably that is from [`yourls_array_granularity()`](https://github.com/YOURLS/YOURLS/blob/b99bbfdc16593d19848cebc9014f1a9d9eda2cc0/includes/functions-infos.php#L304) and not a result of this patch. It's just more likely as the link ages to drop meaningful values from the larger set.)

LeoColomb:
> What worries me a little more is that the click total I get by summing the points in the updated line graph view is only about half the actual total reported in the table to the right.

Can you check if the total in the right table corresponds to the number of:
1. Click count in the database (I expect so)?
2. The number of lines in the database log table for the corresponding link?

As the two might be different, it's eventually all fine.

dgw:
> Can you check if the total in the right table corresponds to the number of:

The total click count is correct. As I said before, it's probably [`yourls_array_granularity()`](https://github.com/YOURLS/YOURLS/blob/b99bbfdc16593d19848cebc9014f1a9d9eda2cc0/includes/functions-infos.php#L304) messing with things to get a "smoother" graph at the expense of accuracy, as it says in the doc comment:

https://github.com/YOURLS/YOURLS/blob/b99bbfdc16593d19848cebc9014f1a9d9eda2cc0/includes/functions-infos.php#L296

Changing that (to e.g. sum the values in each "bucket" instead of dropping the in-betweens) could be an interesting thought, but out of scope for this PR 😸

LeoColomb:
> As I said before, it's probably [`yourls_array_granularity()`](https://github.com/YOURLS/YOURLS/blob/b99bbfdc16593d19848cebc9014f1a9d9eda2cc0/includes/functions-infos.php#L304) messing with things to get a "smoother" graph at the expense of accuracy, as it says in the doc comment

Oh right, apologies (again!), I read your comment too quick.
I see now.

Right, so ready to merge?

dgw (approved):
> Right, so ready to merge?

Yes :shipit:

PR #3910, opened 2025-04-28
Fix "Display 1 to 0 of 0 URLs" on admin list page

SimStim:
You know how it is - you got something perfect, a bleach-white shirt, and people will notice the tiniest speck. This one fixes the message "Display 1 to 0 of 0 URLs." on a new installation. Best regards from your friendly autist. ;-)

### DIFF
--- a/admin/index.php
+++ b/admin/index.php
@@ -242,9 +242,12 @@ yourls_do_action( 'admin_page_before_content' );
 
 if ( !$is_bookmark ) { ?>
 	<p><?php echo $search_sentence; ?></p>
-	<p><?php
+    <p><?php
         if ($total_items === 0)
-            printf( yourls__( 'Displaying <strong>ZERO</strong> URLs, because none have been defined' ));
+        {
+            printf( yourls__( 'No URLs' ));
+            if ($search !== "") printf( yourls__( '. Try being less specific' ));
+        }
         else printf( yourls__( 'Display <strong>%1$s</strong> to <strong class="increment">%2$s</strong> of <strong class="increment">%3$s</strong> URLs' ), $display_on_page, $max_on_page, $total_items );
 		if( $total_items_clicks !== false )
 			echo ", " . sprintf( yourls_n( 'counting <strong>1</strong> click', 'counting <strong>%s</strong> clicks', $total_items_clicks ), yourls_number_format_i18n( $total_items_clicks ) );
--- a/admin/index.php
+++ b/admin/index.php
@@ -242,15 +242,16 @@ yourls_do_action( 'admin_page_before_content' );
 
 if ( !$is_bookmark ) { ?>
 	<p><?php echo $search_sentence; ?></p>
-    <p><?php
-        if ($total_items === 0)
-        {
-            printf( yourls__( 'No URLs' ));
-            if ($search !== "") printf( yourls__( '. Try being less specific' ));
-        }
-        else printf( yourls__( 'Display <strong>%1$s</strong> to <strong class="increment">%2$s</strong> of <strong class="increment">%3$s</strong> URLs' ), $display_on_page, $max_on_page, $total_items );
-		if( $total_items_clicks !== false )
-			echo ", " . sprintf( yourls_n( 'counting <strong>1</strong> click', 'counting <strong>%s</strong> clicks', $total_items_clicks ), yourls_number_format_i18n( $total_items_clicks ) );
+	<p><?php
+		if ( $total_items === 0 ) {
+			printf( yourls__( 'No URLs.' ) );
+			if ( $search !== "" )
+				printf( ' ' . yourls__( 'Try being less specific' ) );
+		} else {
+			printf( yourls__( 'Display <strong>%1$s</strong> to <strong class="increment">%2$s</strong> of <strong class="increment">%3$s</strong> URLs' ), $display_on_page, $max_on_page, $total_items );
+			if( $total_items_clicks !== false )
+				echo ", " . sprintf( yourls_n( 'counting <strong>1</strong> click', 'counting <strong>%s</strong> clicks', $total_items_clicks ), yourls_number_format_i18n( $total_items_clicks ) );
+		}
 	?>.</p>
 <?php } ?>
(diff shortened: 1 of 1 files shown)


ozh:
Nice catch, never noticed. This said, you hardcoded the message, making it untranslatable. See https://yourls.org/docs/development/i18n for how to make it

SimStim:
> Nice catch, never noticed. This said, you hardcoded the message, making it untranslatable. See https://yourls.org/docs/development/i18n for how to make it

OMFG you got me with my pants down (this time!) I'm very sorry for the oversight, and I hope you can find it in the bottom of your heart to forgive me.

Of course it should take into account whatever i18n technique you're using. I acted rashly and was still in rage mode earlier. That said, I guess your reply means you would be rather inclined to accepting the pull if I put in some real work. :D Right now I'm on my way out, but will look into everything properly later on, cheers!

SimStim (commented):
I cannot find any .po or .mo files in the project. I guess this is just "prepared" for l10n, but there's only one language as of now? Or maybe I'm missing something..
Well, I inserted that one line I changed here into my local installation via VI and it works :) Take it or leave it.

admin/index.php
```diff
@@ -243,7 +243,9 @@
 if ( !$is_bookmark ) { ?>
 	<p><?php echo $search_sentence; ?></p>
 	<p><?php
-		printf( yourls__( 'Display <strong>%1$s</strong> to <strong class="increment">%2$s</strong> of <strong class="increment">%3$s</strong> URLs' ), $display_on_page, $max_on_page, $total_items );
+        if ($total_items === 0)
+            printf( yourls__( 'Displaying <strong>ZERO</strong> URLs, because none have been defined' ));
```
dgw, line 247: This can also appear if the user applied a filter with no matching results. "because none have been defined" is only accurate for one specific case.

```suggestion
            printf( yourls__( 'Displaying <strong>0</strong> URLs matching the current filters' ) );
```

Perhaps there's a better way to phrase it, or if you think it's worth the effort you could add a check if any filter conditions are set and display a different message.

dgw (commented):
> Nice catch, never noticed.

The DB is initialized with a couple example links, so someone has to intentionally bring a new install down to zero stored items in order to see this :)

> I cannot find any .po or .mo files in the project. I guess this is just "prepared" for l10n, but there's only one language as of now? Or maybe I'm missing something..

A `.pot` is maintained automatically in a separate repo: https://github.com/YOURLS/YOURLS.pot

You are correct that YOURLS itself contains only English localization. Translations based on the `.pot` are submitted by their maintainers to be listed on the Awesome List: https://github.com/YOURLS/awesome#translations

admin/index.php
SimStim: Thank you for the additional information. First time user here, I have less than 10 links in my setup and not using filters yet...

So, my proposal: let's use a wording that is universal while still useful. I'm thinking: "no URLs to display." Would that work for you?

dgw: I've come to really like the idea of indicating to the user that filters are active, but "No URLs to display" does work! Short, sweet, and simple.

SimStim: Well all right then, that decision is totally up to you. It is called a pull request not coercion. Requests can be denied! 😊

SimStim: No I'm not sure I understood your comment properly... Just wanted to say anyway you decide it's fine with me. Peace.

dgw: Go with a simple "No URLs to display" for this PR.

If I feel strongly enough about having a different message when non-default filters are active, I can make another patch myself later :slightly_smiling_face:

SimStim: Let me just take a moment out of our busy days and appreciate the existence of actively maintained projects on GitHub. There are too many "other ones" around. I will make the changes as discussed once I'm awake!

dgw:
You know, I was doing my level best to make a code review suggestion that would tweak all the style points I wanted to hit (tabs instead of spaces, whitespace around parameters, our little conventions) and then realized I should test the thing.

Testing (live, in production 😱) got me thinking about more nuanced logic for display of different message parts, so I hope you don't mind that I just ended up authoring a "code review commit" directly applying the changes I'd make. Try it out! I think the output works very nicely in all cases.\*

----

\* — I didn't delete all URLs from my instance to test that case, but I hit "search with results", "search with 0 results", and the default list without any filters.

SimStim:
Sorry about the tabs and spaces, I had no idea you feel so strongly about it. You see I don't have an IDE at the moment. I edited the index.php on my server with VI, then copy pasted from the bash console into GitHub. No wonder the tabs got messed up. 

As for testing, you don't need to remove your urls.  Just put $total_items = 0 before the if and you're good to test, even on a live system, because it doesn't affect the redirection functionality.

dgw (approved):
Will let another maintainer look this over (including my tweaks).

> Just put $total_items = 0 before the if and you're good to test

Ah of course, I'm clearly not in "webapp brain" today. That case is now tested too.

> You see I don't have an IDE at the moment. I edited the index.php on my server with VI, then copy pasted from the bash console into GitHub. No wonder the tabs got messed up.

Funny enough, I was initially making tweaks on _my_ server in VI. No, I don't trust its automatic tabs/spaces; final patch was checked in VS Code locally before committing.

admin/index.php
```diff
@@ -243,9 +243,15 @@
 if ( !$is_bookmark ) { ?>
 	<p><?php echo $search_sentence; ?></p>
 	<p><?php
-		printf( yourls__( 'Display <strong>%1$s</strong> to <strong class="increment">%2$s</strong> of <strong class="increment">%3$s</strong> URLs' ), $display_on_page, $max_on_page, $total_items );
-		if( $total_items_clicks !== false )
-			echo ", " . sprintf( yourls_n( 'counting <strong>1</strong> click', 'counting <strong>%s</strong> clicks', $total_items_clicks ), yourls_number_format_i18n( $total_items_clicks ) );
+		if ( $total_items === 0 ) {
+			printf( yourls__( 'No URLs.' ) );
+			if ( $search !== "" )
```
LeoColomb, line 248:
```suggestion
			if ( ! empty( $search ) )
```

LeoColomb (approved):
Despite a tiny tweak, lgtm

PR #3911, opened 2025-04-30
Add __wakeup() method to Iri class to prevent PHP object injection vulnerabilities

lexo-mfleuti:
# Security Fix: Prevent PHP Object Injection in Iri Class

## Description
This pull request adds the `__wakeup()` magic method to the `Iri` class to address a potential PHP object injection vulnerability during unserialization operations.

## Vulnerability Details
When PHP unserializes objects, the `__wakeup()` method is called automatically. Without proper validation in this method, an attacker could craft malicious serialized strings that, when unserialized, might:
- Manipulate object properties in unexpected ways
- Cause type confusion issues
- Potentially lead to code execution in certain contexts
- Create vectors for property-based attacks

## Changes Made
The PR adds:
- A new `__wakeup()` method that performs type checking on all class properties
- Validation to ensure string properties remain strings
- Validation to ensure array properties remain arrays
- Exception throwing for invalid types
- Property resetting as an additional safety measure

## Testing
The changes maintain backward compatibility while adding security controls. Standard unit tests pass with this change.

## Impact
This improvement prevents object injection attacks when deserializing `Iri` objects from untrusted sources, strengthening the overall security posture of the library.

## References
- [[OWASP: PHP Object Injection](https://owasp.org/www-community/vulnerabilities/PHP_Object_Injection)](https://owasp.org/www-community/vulnerabilities/PHP_Object_Injection)
- [[PHP: Magic Methods Documentation](https://www.php.net/manual/en/language.oop5.magic.php)](https://www.php.net/manual/en/language.oop5.magic.php)

### DIFF
--- a/includes/vendor/rmccue/requests/src/Iri.php
+++ b/includes/vendor/rmccue/requests/src/Iri.php
@@ -717,6 +717,38 @@ class Iri {
 		return true;
 	}
 
+    /**
+    * This method was added to prevent object injection vulnerabilities during PHP object unserialization.
+    * 
+    * When PHP unserializes objects, the __wakeup() magic method is called automatically. Without
+    * proper validation in this method, attackers could potentially craft malicious serialized strings
+    * that, when unserialized, could lead to unexpected behavior, data exposure, or remote code execution.
+    * 
+    * The implemented fix validates that class properties are of expected types after unserialization:
+    * - String properties (scheme, iuserinfo, ihost, port, ipath, iquery, ifragment) must be strings
+    * - Array properties (normalization) must be arrays
+    * 
+    * If any property has an unexpected type, an UnexpectedValueException is thrown, preventing the 
+    * potentially dangerous object from being fully instantiated and used in the application.
+    * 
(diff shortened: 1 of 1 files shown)


dgw:
Files under `include/vendor` are managed by Composer, and will get overwritten next time those dependencies are synced.

Better to raise this upstream.

lexo-mfleuti:
OK, however you'd like to handle it. But it should find its way into production.

LeoColomb:
Not only it should be done upstream, but a vulnerability fix should probably not be publicly published like that, especially when:
* There has been not vulnerability reported prior to the PR (nor an external reference linked)
* The pull request is clearly made using AI extensively

Closing

lexo-mfleuti:
Our providers' security tool has identified the issue and fixed it. I used AI to quickly create the necessary descriptions and commit messages with minmal effort since I don't have much time but wanted to help. Hence I tried to publish an easy solution for you. I fully agree with your statement/reasoning. There'd be smarter ways to do that ;)

Screenshot from our providers security tool: https://screenshare.lexo.ch/2025-05-01-191452.png

PR #3929, opened 2025-06-07
Update readme.html : remove Gandi

ozh:
Remove Gandi.net - those folks have gotten insanely expensive and are no longer the cool alternative registrar they used to be.


LeoColomb (approved):
I do so too

PR #3988, opened 2025-09-20
Replace tabs

ozh:
Replacing all remaining tabs to spaces and make sure it's blame-ignored

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,2 +1,5 @@
 # End of line reformating
 3daf6593d89c608a6660a6c0b872eeb2607548ba
+# Convert all tabs to spaces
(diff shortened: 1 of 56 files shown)


dgw (approved):
I'm pleasantly surprised that viewing blame on the head commit's version of one of the files took the updated blame-ignore into account. GH could totally say "only the file on your default branch counts", but it didn't. So I've already tested that these whitespace changes get ignored!

ozh:
Actually, is it working as expected ?
image

dgw:
There is no other content. I guess _some_ commit has to take the blame for each line, so even an ignored commit can appear if there's no "real" changes to show. 🤷‍♂️

PR #3997, opened 2025-10-10
Fix case sensitivity in search

ozh:
Stance : search should be CASE SENSITIVE regarding URLs and short URLs, but CASE INSENSITIVE regarding titles.

### DIFF
--- a/admin/index.php
+++ b/admin/index.php
@@ -34,13 +34,16 @@ if( $search && $search_in && $search_in_text ) {
     $search_text     = $search;
     $search          = str_replace( '*', '%', '*' . $search . '*' );
     if( $search_in == 'all' ) {
-        $where['sql'] .= " AND CONCAT_WS('',`keyword`,`url`,`title`,`ip`) LIKE (:search)";
-        // Search across all fields. The resulting SQL will be something like:
-        // SELECT * FROM `yourls_url` WHERE CONCAT_WS('',`keyword`,`url`,`title`,`ip`) LIKE ("%ozh%")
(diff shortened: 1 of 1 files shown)


ozh:
Testing with https://ozh.org/temp/UPLOL-lowrolf.html : 

✔️ case sensitive for URL
image
image
image
image


✔️ case sensitive for short URL
image
image
image


✔️ case **in**sensitive for titles
image
image
image
image

dgw:
I tweaked https://github.com/YOURLS/YOURLS/pull/3997#issuecomment-3391689082 a bit so it accurately describes the behavior of titles. :)

Is it overkill to use `COLLATE` now that you replaced `CONCAT_WS()` with multiple <code>\`field\` LIKE (:search)</code> clauses? If tested individually, I would think that each column would naturally behave according to its set collation. For `title` + `ip`, that's already `utf8mb4_unicode_ci`, and `utf8mb4_bin` for `url` + `keyword`.

If we can accomplish the same thing with fewer conditionals, it's nicer—and probably more performant if we can avoid making the DB convert collations. (Not that I expect search to be a major performance issue, but… why make things slow just because? 😁)

ozh:
With no `COLLATE` clause, it doesn't work : 

```php
        $where['sql'] .= " AND `keyword` LIKE (:search)
                        OR `url` LIKE (:search)
                        OR `title` LIKE (:search)
                        OR `ip` LIKE (:search) ";
```
image

dgw:
Ah, fun. Guess we'll do it this way, indeed.

Issue #3802, opened 2024-08-31
Slashes is stripped


jcubic:
### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/YOURLS/.github/blob/main/CODE_OF_CONDUCT.md).

### Submission validity

- [X] This is not a personal support request, that should be posted on [discussions community](https://github.com/YOURLS/YOURLS/discussions).
- [X] I checked [current issues](https://github.com/YOURLS/YOURLS/issues) and this request isn't a duplicate of an existing issue, opened or closed.

### Self troubleshooting

- [X] I've read the [documentation](https://yourls.org/) and made sure to follow it.
- [X] I've read the [troubleshooting first steps](https://docs.yourls.org/guide/troubleshooting/first-steps.html) and [frequent issues](https://docs.yourls.org/guide/troubleshooting/frequent-issues.html) guides.

### Version

1.9.2

### Description

When shortening the ULR that contain slashes YOURLS strip them when saving to database.

### Expectation

I have application that contain JSON in URL hash if the strings contain quotations the ULR doesn't work when using shortener. Because `"` inside string in JSON become `\"` and YOURLS saves `"` and you end up with invalid JSON.

### Reproduction steps

1. Add this URL: `https://terminal.jcubic.pl/404#[[0,1,%22jargon%20\%22Don't%20do%20that%20then!\%22%22]]` that should run a command in the terminal
2. Open the short link
3. The command don't appear (the web app ignores any JSON errors)

### Context

I think that the tool should not do any manipulation to the URL hash, it should always be passed as is. Of course, with escapes and unescapes when writing and reading from database.

Browser do some escaping when copying full URL, here is same URL with hash copied separately, so there are no escaping:

```
https://terminal.jcubic.pl/404#[[0,1,"jargon%20\"Don't%20do%20that%20then!\""]]
```

As you can see, there is `\"` in the hash. When copying this URL quotations and backsplashed are removed.

PR #3998, opened 2025-10-11
Sanitize: preserve backslashes in URLs

matt-h:
This is a fix for #3802

This update preserves backslashes in URLs

The sample from the issue of `https://terminal.jcubic.pl/404#[[0,1,%22jargon%20\%22Don't%20do%20that%20then!\%22%22]]` now works.

Added regression tests for #3802 
Also added a test to make sure that a more generic `https://example.com/path?q=a\b\c#x\y\z` works as expected.

### DIFF
--- a/includes/functions-formatting.php
+++ b/includes/functions-formatting.php
@@ -539,6 +539,18 @@ function yourls_esc_url( $url, $context = 'display', $protocols = array() ) {
     $url = yourls_normalize_uri( $url );
 
     $url = preg_replace( '|[^a-z0-9-~+_.?#=!&;,/:%@$\|*\'()\[\]\\\\\\x80-\\xff]|i', '', $url );
+    // The replace above allows backslashes now, but we only should only allow them after a query string or a fragment identifier.
+    if ( str_contains( $url, '\\') ) {
+        if ( str_contains( $url, '?') ) {
+            $parts = explode( '?', $url );
+            $url = str_replace( '\\', '', array_shift( $parts ) ) . '?' . implode( '?', $parts );
+        } elseif ( str_contains( $url, '#') ) {
+            $parts = explode( '#', $url );
+            $url = str_replace( '\\', '', array_shift( $parts ) ) . '#' . implode( '#', $parts );
+        } else {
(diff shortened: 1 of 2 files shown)


ozh:
Thanks for this proposal !

Slightly broader question : should we allow backslashes in URL (or more generally in any URI), or just in the query string or fragment, as in your use case ?

matt-h:
> Slightly broader question : should we allow backslashes in URL (or more generally in any URI), or just in the query string or fragment, as in your use case ?

If you try to put backslashes in a url in the browser it just swaps them for forward slashes. They are only allowed in query strings and fragments.

ozh:
The problem isn't just how Firefox or Chrome behave, it's also in inner parts of YOURLS for instance when we fetch a page to get its title. Backslashes immediately raise a warning in my mind about directory traversal or "creative" stuff like this

matt-h:
That makes sense, but since they aren't valid to visit in a browser for a redirect it shouldn't allow the to be put in.
I can add some more tests to make sure that they are being sanitized out properly if that makes sense to you?

ozh:
The thing is, there are people storing much more than links in YOURLS (we've had users storing data uri, local files, emails, so who knows what else)

Why not adopt a conservative and safe approach and restrict backslashes to URL query strings and fragments only ?

matt-h:
I see, I was misunderstanding the regex and I was thinking it was already working properly that way. After more testing I see that it is now allowing backslashes anywhere.

I'm putting in more tests to catch these cases and will update it again to allow in the query parameters and fragments while not allowing anything that currently isn't allowed.

matt-h:
@ozh I have updated this now so it still removes backslashes before the question mark or hash in a url, the same behavior as before.

Tests are updated to check the different cases of urls formatted this way.

ozh:
Doing an `if contains '?' elseif contains '#'` didn't account for https://example.com/path#x\\y\\z?hello\\world (we can have both `?this#that` and `#this?that`

I went a simpler way

In the process I fucked something wrong because it created a new PR instead of completing yours, sorry for the mess 😢 -> https://github.com/YOURLS/YOURLS/pull/4000

PR #4005, opened 2025-10-24
hide referrers on public statistics page by default

### LINKED ISSUE #3670: By default, hide referrers on public stats
By default we're showing referrers on public stat pages -- example : https://yourls.org/ozh+ see on "Traffic Sources"
To prevent any 2000's old school referrer spam, or maybe for privacy, we should display these only if the stat pages are private

So, if `yourls_is_private() !== true`, don't display the whole "Traffic Sources" tab. Make this filterable so that anyone absolutely willing to display if can do

image

matt-h:
add `statistics_show_referrers` filter to override.

Fixes #3670

### DIFF
--- a/yourls-infos.php
+++ b/yourls-infos.php
@@ -209,6 +209,10 @@ if( yourls_do_log_redirect() ) {
 
 }
 
+// Whether to show referrers on the statistics page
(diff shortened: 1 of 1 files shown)


matt-h:
Which constant is that?

PR #4008, opened 2025-10-25
Set the infos cache of the keyword after a url is added/updated/deleted

### LINKED ISSUE #3772: YDB cache array not updated after new link creation
### Code of Conduct

### Submission validity

### Self troubleshooting

### Version

1.9.2

### Description

In my bootstrap setup, I use yourls_get_keyword_infos() to first check if a keyword exists and then, if the keyword doesn't exist, I use yourls_add_new_link() to insert the link.

My use of yourls_get_keyword_infos() before creation results in "FALSE" going in the YDB cache array for the keyword. Then, since yourls_add_new_link() does not update the YDB cache array after creation, the admin interface incorrectly reports that the newly created keyword is not in the database.

After the creation of a new link, the entry for the keyword in the YDB cache array should be removed so it can properly be cached on the next lookup.

### Expectation

_No response_

### Reproduction steps

_No response_

### Context

_No response_

matt-h:
Set the infos cache of the keyword after a url is added/updated/deleted to/from the database.

This will allow it to overwrite the cache with the updated keyword if a cache was already set on the request. Basically it will keep the cache in sync with changes as they happen. Purging the cache in the case it doesn't know what to update.

This will prevent plugins accessing cached data from accessing stale data.

Adds a new function `update_infos_if_exists` to the YDB class to update values on an existing cache if it exists. Basically a helper to reduce redundancy on updating cache.

Add tests to test the logic.

Fixes #3772

### DIFF
--- a/.github/workflows/update-translations.yml
+++ b/.github/workflows/update-translations.yml
@@ -49,7 +49,7 @@ jobs:
             --keyword=yourls_n_noop:1,2 \
             --keyword=yourls_nx_noop:1,2
 
-      - uses: actions/upload-artifact@v4
+      - uses: actions/upload-artifact@v5
         with:
           name: YOURLS-pot
           path: YOURLS.pot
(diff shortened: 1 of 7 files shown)


ozh (approved):
Fantastic once again ❤️

PR #4013, opened 2025-11-02
Add a url index for faster SQL lookups with large tables

### LINKED ISSUE #3793: `YOURLS_UNIQUE_URLS` slows down large instances

matt-h:
This is a replacement for the hash index I added in https://github.com/YOURLS/YOURLS/pull/4006

I found that just adding a basic index to the urls column limited to 30 chars has the same performance boost as the url hash. This handles it all in mysql without adding any new complexity to the codebase or queries.

To note, when doing a lookup by the index it will match on those first 30 characters and to filter it down to matching ones and then do a WHERE on just those results which is significantly faster than doing a WHERE on the entire table. The one drawback vs the hash would be the case where every url in the database is the same domain name that is longer than 30 characters (not a likely case in my opinion) it would just work the same as it does now.

For data size difference, my test DB with over 4 million urls was 1.4 GB with the current version of YOURLS, 1.6 GB with the hash version and also 1.6 GB with this new index version.

Also, the upgrade time was about 1/5 the time of adding the hash in my testing.

Fixes #3793

### DIFF
--- a/includes/functions-api.php
+++ b/includes/functions-api.php
@@ -124,8 +124,14 @@ function yourls_api_output( $mode, $output, $send_headers = true, $echo = true )
             if( $send_headers )
                 yourls_content_type_header( 'application/javascript' );
 
-            $callback = isset( $output['callback'] ) ? $output['callback'] : '';
-            $result =  $callback . '(' . json_encode( $output ) . ')';
+            $callback = isset( $output['callback'] ) ? yourls_validate_jsonp_callback($output['callback'] ) : '';
+            if( $callback === false ) {
+                yourls_status_header( 400 );
+                $result = json_encode( ['errorCode' => '400', 'error' => 'Invalid callback parameter'] );
+            } else {
+                $result =  $callback . '(' . json_encode( $output ) . ')';
+            }
+
             break;
 
         case 'json':
--- a/includes/functions-formatting.php
+++ b/includes/functions-formatting.php
@@ -265,6 +265,39 @@ function yourls_sanitize_filename($file) {
 }
 
 /**
+ * Validate a JSONP callback name
+ *
+ * Check if the callback contains only safe characters: [a-zA-Z0-9_$.]
+ * Returns the original callback if valid, or false if invalid.
+ *
+ *  Examples:
+ *  - 'myCallback' => 'myCallback'
+ *  - 'alert(1)'   => false
+ *  See tests/tests/format/JsonpCallbackTest.php for various cases covered
+ *
+ * @since 1.10.3
+ * @param string $callback Raw callback value
+ * @return string|false Original callback if valid, false otherwise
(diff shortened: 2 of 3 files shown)


ozh:
Smart, thanks, indeed much better

dgw:
It happens :)
That's why we have multiple people look at these things!

joesenova:
Glad this is sorted 4 years after I recommended the fix:

https://github.com/YOURLS/YOURLS/pull/2585

ozh:
For the record this is NOT the fix your recommended @joesenova . To be accurate, this is more the suggestion I made to you 4 years ago, but to which you didn't reply.

joesenova:
Each to their own @ozh .

Glad to see that this issue is fixed, thanks for that.

2ndK16:
I'm not sure how urgent this is, but there are different index lengths in `function-install.php` (30) and ` `function-upgrade.php` (50).

ozh:
Oops, nice catch @2ndK16   
I propose we update the `functions-install.php` to 50 and leave the `functions-upgrade.php` as is, to avoid a database upgrade for everyone

Issue #3846, opened 2025-01-15
Allow for Read Write splitting mysql querries


daynesuke:
### Proposed functionality

For those who want it, allows the functions with read only queries to be sent to a read only MySQL instance. 

A few example functions are:

- yourls_get_db_stats
- yourls_get_link_stats
- yourls_get_stats

### Use case

**Why are we doing this?**

We are doing this to lower the traffic load on the write/master MySQL installation. By sending the read only queries to another MySQL installation it lowers the overall load on the write/master allowing it to focus on doing only the writes.

**What use cases does it support?**

A use case for this would be when the DB has millions of URLs added to it, all of the unique checks and other SELECT statements run when creating a new URL can over load the system if there are lots of requests for new links coming in. By shunting the read queries to another MySQL instance we can alleviate the stress put on the write/master.

Another use case would be to fully utilize AWS Aurora MySQL installations, which have a master and read only instance in their cluster by default.

A final use case would be to continue serving traffic for existing links when the master/write instance is overloaded with writes, or unavailable for any other reason. This would only be applicable of clicks tracking is disabled, or being handled differently.

**What is the expected outcome?**

The expected outcome is that YOURLS can handle more reads and writes on big and/or busy installations.

### Potential solution

What I propose is to add some new sections to the config.php file for setting up a read only connection. Then in the class-mysql.php add another variable/property called $rdb (or whatever else you want to call it) that establishes a connection with the read only settings. Finally, update all of the functions in functions.php that have SELECT only queries to use $rdb instead of $ydb.

An alternative I was considering was to make my own plugin for this. However, I came to realize that many of the read only functions do not have the shunt option to allow plugins to override them (see the Proposed functionality section for a few examples). I debated on adding the shunt option to the ones that are missing, then realized if I was going to do that I should just change the $ydb to $rdb. However, since it is heavily [advised ](https://yourls.org/docs/development/dont-hack-core) that changing the Core files should not be done, I figured I would put in a feature request as others might be interested in something like this.

I started looking into this when our RDS installation was overloaded with new URL requests. Investigating that led me to find a bunch of Select queries that we could have run on a Read Replica. In our specific case we are allowing for duplicate URLs and have disabled click tracking so we would like to be able to serve links from a Read Replica as well.

ozh:
I like this idea of being able to use different SQL connections.

If we just add an action to `yourls_get_db()`, I think it's possible with a simple plugin to delegate each individual query to a chosen DB connection.

```php
[...]

PR #4020, opened 2025-11-23
Add context to `yourls_get_db()`

ozh:
PR related to #3846 

All `yourls_get_db()` calls in core are now passing an optional `$context` parameter, with the following naming convention: 

```
/*
 * @param string $context Optional context. Default: ''.
 *   When provided, use a naming schema starting with a prefix describing the operation, followed
 *   by a short description
 *   - Prefix should be one of: "read-", "write-", or "other-".
 *        * "read-" for operations that only read from the DB (eg get_keyword_infos)
 *        * "write-" for operations that write to the DB (eg insert_link_in_db)
 *        * "other-" for operations that do not fit in the other categories because they are not SQL queries (eg
 *           when just storing info in memory, in the YDB object)
 *   - The description should be lowercase, words separated with underscores, eg "insert_link_in_db".
 *   Examples:
 *   - read-fetch_keyword
 *   - write-insert_link_in_db
 *   - write-update_url_title
 *   - read-get_options
 *   - other-get_debug_log
```

One practical use of such contextualized `yourls_get_db()` calls is to be able to use different DB connections in a Master/Slave setup. The following plugin should be a helping start : https://gist.github.com/ozh/553c2519878984271584939355e91b8e

@daynesuke Could you test this branch and the plugin for your use case ? Input would be much appreciated

### DIFF
--- a/includes/geo/COPYRIGHT.txt
+++ b/includes/geo/COPYRIGHT.txt
@@ -1,1 +1,1 @@
-Database and Contents Copyright (c) 2025 MaxMind, Inc.
+Database and Contents Copyright (c) 2026 MaxMind, Inc.


includes/functions-upgrade.php
```diff
@@ -403,7 +403,7 @@ function yourls_alter_url_table_to_14_part_two() {
  *
  */
 function yourls_update_table_to_14() {
-    $ydb = yourls_get_db();
+    $ydb = yourls_get_db('read-update_table_to_14');
```
dgw: The potential `UPDATE` queries below are pretty `write-`y.

```suggestion
    $ydb = yourls_get_db('write-update_table_to_14');
```

Note also L428 has a couple typos that I can't make suggestions for ("Huho" -> "Uh-oh"; "rown" -> "row") since they're too far from the patch. (Dumb, arbitrary limitation IMO.)

tests/tests/db/DBSetTest.php
```diff
@@ -40,10 +46,13 @@ public function test_set() {
...
-     *     glȍbal $ydb;
-     *     $this->assertTrue( isset($ydb) );
-     *     yourls_set_db(null);
-     *     $this->assertFalse( isset($ydb) );
+     *     $db = yourls_get_db();
+     *     $this->assertTrue( isset($db) );    // OK
+     *
+     *     yourls_set_db(null);                // should unset $ydb
+     *     global $ydb;
+     *     yourls_ut_var_dump( $ydb);          // $ydb is still set and has the same value
```
dgw: Nitpick:

```suggestion
     *     yourls_ut_var_dump( $ydb );         // $ydb is still set and has the same value
```

tests/tests/db/DBSetTest.php
```diff
@@ -32,6 +32,12 @@ public function test_set() {
         $this->assertSame( "hello", yourls_get_db() );
     }
 
+    public function test_context_optional_parameter() {
+        $db_default = yourls_get_db();
+        $db_with_ctx = yourls_get_db('some_context');
+        $this->assertSame($db_default, $db_with_ctx);
+    }
```
dgw: Consider testing with a valid prefix as described in the docstring for `yourls_get_db()`, and adding a separate test with an "invalid" context value.

includes/class-mysql.php
```diff
@@ -76,18 +81,34 @@ function yourls_db_connect() {
...
  *
  * @since  1.7.10
+ * @param string $context Optional context. Default: ''.
+ *   When provided, use a naming schema starting with a prefix describing the operation, followed by a short description:
+ *   - Prefix should be one of: "read-", "write-", or "other-".
+ *        * "read-" for operations that only read from the DB (eg get_keyword_infos)
+ *        * "write-" for operations that write to the DB (eg insert_link_in_db)
+ *        * "other-" for operations that do not fit in the other categories because they are not SQL queries (eg
+ *           when just storing info in memory, in the YDB object)
+ *   - The description should be lowercase, words separated with underscores, eg "insert_link_in_db".
```
dgw: I would suggest validating the prefix somehow, or at least explicitly specifying how a `$context` that doesn't fit this naming scheme should be handled.

ozh:
All good points here @dgw , thanks a lot.

Suggestion for the `read-` / `write-` / `other-` situation : drop the `other-` and have just `read-` and `write-`. Other cases when it's not strictly a read or write statement can be finely handled by the slave DB if any.

tests/tests/db/DBSetTest.php
dgw: That's what I mean. You define a naming schema when documenting the `context` parameter in `class-mysql.php`:

```
 * @param string $context Optional context. Default: ''.
 *   When provided, use a naming schema starting with a prefix describing the operation, followed by a short description:
 *   - Prefix should be one of: "read-", "write-", or "other-".
 *        * "read-" for operations that only read from the DB (eg get_keyword_infos)
 *        * "write-" for operations that write to the DB (eg insert_link_in_db)
 *        * "other-" for operations that do not fit in the other categories because they are not SQL queries (eg
 *           when just storing info in memory, in the YDB object)
 *   - The description should be lowercase, words separated with underscores, eg "insert_link_in_db".
```

To me that creates an expectation for plugins that if there is a `$context`, it will match this format.

dgw:
I guess that "slave DB" means `read-`? That's fine, as long as the `other-` it's replacing doesn't actually write to the DB. (Think I flagged some of those that you updated in the PR as-is, around plugin activation maybe?)

If the primary originator of `$context` parameters is _us_, i.e. core devs, we should be OK. :) I'll look for the patch that replaces `other-`s with one of the other two.

ozh:
Calling `yourls_get_db()` with an improperly formatted context now adds a debug message in the log: 

image

ozh:
Morning thought : maybe instead of message in the footer, we could raise a notice, like we do [when someone uses a deprecated function](https://github.com/YOURLS/YOURLS/blob/master/includes/functions.php#L1260-L1271). Notices show only if YOURLS_DEBUG is true so this change would not affect YOURLS instances in prod

ozh:
What do we prefer ? When YOURLS_DEBUG is true,

Information message in footer : 
image
(which should contain file and line num to be more useful)

or plain old PHP notice : 
image

Pros and cons : 
- we, core devs, might forget to add a context to future `yourls_get_db()` calls with a footer message, not with a notice since PHPUnit screams
- with a real notice, the context is "less optional", while it will really benefit to 0,01% of users when they'll use a master/slave setup (there might be another future use for this context, let's not rule this out completely)

dgw:
100% go with the approach that has the best chance of making PHPUnit complain, imo

I believe `display_errors` is often set to Off by default since it can expose webserver stuff, so it might make sense to emit _both_ a notice and a footer message. 🤷‍♂️

ozh:
Missing or malformed contexts now trigger a notice. No message in the footer for consistency with other deprecation messages in YOURLS

LeoColomb:
Principle sounds all good to me.
One question, though: in case the context is missing, is it well assumed the request is `write` as a fallback?
For existing plugins doing database stuff, that would help avoid split issues.

ozh:
If the context is omitted, a notice is thrown, but nothing else happens.

As I see it, someone making a plugin for a master/slave setup should make it so that, by default, unspecified queries are sent to the `write` connection, in order to deal with existing code using `yourls_get_db()->do_something()` or, very common, `$ydb->do_something()`.

Maybe I should stress this out in my sample plugin

LeoColomb (approved):
Understood. I think I would have preferred an internal fallback, but all good for me both ways.

PR #4025, opened 2025-12-07
Fix hover to display links in tables for users with screen readers.

lilmike:
The details, delete, etc links in tables were hidden with display: none until hovered, which hid them from screen readers. Screen readers do not hover the columns with the links, however, so we need to hide them in a way that keeps them available for screen readers.

Please do tell me if any of the css is wrong, or if the web interface looks wrong visually -- I am also blind, so I don't know much about css and cannot tell if things look wrong.

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -129,23 +129,32 @@ tr.edit-row  td {
 td.url small a{
 	color:#bbc;
 }
-body.desktop td.actions input,body.desktop  td.actions a {
-	position: absolute !important;
-	left: -9999px;
-	clip: rect(1px, 1px, 1px, 1px);
+/* Hide buttons visually but keep them accessible to screen readers */
+body.desktop td.actions input, body.desktop td.actions a {
+	opacity: 0;
+	pointer-events: none;
+	transition: opacity 0.2s ease;
 }
 td.timestamp span.timestamp {
     display:none;
 }
 td.actions input.disabled, td.actions input.loading {
-	position: relative !important;
-	left: auto;
-	clip: auto;
+	visibility:visible;
 }
+/* Show buttons on row hover */
 tr:hover td.actions input, tr:hover td.actions a {
-	position: relative !important;
-	left: auto;
-clip: auto;
+	opacity: 1;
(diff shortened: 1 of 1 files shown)


ozh:
EDIT : oops nope, after I refreshed cache :-P

It doesn't work : the icons are now either hidden or piled, but it's unusable as of now

lilmike:
@ozh Does this happen when hovering or when not hovering the icons? Knowing
this will allow me to hopefully figure out which css statement needs
changing, though I suspect its the one for when icons are hovered.

-Michael.






-- 
Sorry about the test, it's just a me thing.
Michael Taboada, Creator of Games. Does that make me a god, well, sorta.
My pgp key: 79BC390E8864CD71305D4DD606CD2197A6EE5ED7
My website: https://michaels.world
2MB website: https://2mb.games
Don't believe everything you read on the internet. I might not be real, you might not be real, and this email certainly isn't real.
Sent from my musically gifted...
stationary bicycle

ozh:
They appear all stacked up since only 1 weird icon shows. Hovering doesn't not change anything

lilmike:
@ozh Try the update I just made... I think not including !important on
the hover section was breaking things. I also added a push left of
-9999px, which should make sure they are out of the visible area.

-Michael.






-- 
Sorry about the test, it's just a me thing.
Michael Taboada, Creator of Games. Does that make me a god, well, sorta.
My pgp key: 79BC390E8864CD71305D4DD606CD2197A6EE5ED7
My website: https://michaels.world
2MB website: https://2mb.games
Don't believe everything you read on the internet. I might not be real, you might not be real, and this email certainly isn't real.
Sent from my supernatural...
stationary bicycle

ozh:
Still doesn't work. Icons are invisible and shown on hover as intended, but the showing makes the table resize on the fly.

I'll come up with a solution.

ozh:
@lilmike can you test this

lilmike:
@ozh It does work just fine for my screen reader. Consider it good.

-Michael.

PR #4030, opened 2025-12-13
Validate jsonp callbacks, add unit tests

ozh:


ozh:
> the calling script won't do what's expected. Might as well make the reason obvious, IMO

Absolutely. Will change this 👍

PR #4036, opened 2025-12-23
Add PHP versions 8.5 and 8.6

ozh:
Of course this will fail, I always manage to break those yml files

### DIFF
--- a/admin/index.php
+++ b/admin/index.php
@@ -326,6 +326,8 @@ yourls_table_end();
 
 yourls_do_action( 'admin_page_after_table' );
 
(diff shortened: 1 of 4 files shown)


ozh:
LGTM, ping @dgw & @LeoColomb

ozh:
(do we like, or need, `--testdox` ?)

dgw (approved):
Regarding `--testdox` I'm neutral. Human readable output is nice, I guess, but so is having the exact test name for easy ctrl-f in the source. 🤷‍♂️

PR #4037, opened 2025-12-23
Update spatie/array-to-xml

ozh:
`composer require spatie/array-to-xml --ignore-platform-reqs` to bypass `aura/sql 6.0.0 requires php ^8.4 -> your php version (8.1; overridden via config.platform, actual: 8.4.5) does not satisfy that requirement`

This lib was throwing PHP deprecations on tests  with 8.4+

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,8 +2,14 @@ name: CI
 
 on:
   push:
-  pull_request:
     branches: [ master ]
+  pull_request:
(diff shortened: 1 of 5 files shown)


ozh:
Merge #4036 first then review here

dgw:
I saw that #4036 was supposed to come first, so I went there. Will update this branch so CI can re-run on the newer PHP versions, then hopefully find a moment to review this PR in the next couple days.

PR #4042, opened 2026-01-02
Update tests

ozh:
- remove `--testdox`, because, seriously, readability
- remove deprecated (and previously unused anyway) `@uses` tags
- restore `DomainTest::test_custom_fake_domain()` and catch the expected notice
- remove all `PHP_VERSION_ID` checking for PHP < 8


LeoColomb:
> * remove `--testdox`, because, seriously, readability

Yes, but it brings context on CI runs (testdox is relatively universal in such consoles).

Issue #4045, opened 2026-01-07
Resonsive web redesign


LeoColomb:
### Proposed functionality

From https://github.com/YOURLS/YOURLS/issues/2693#issuecomment-3383452310 by @Aternus

> I'm using YOURLS on a daily basis from my phone and would like to update the theme, is this still relevant or is somebody is already working on the issue?

### Potential solution

From https://github.com/YOURLS/YOURLS/issues/2693#issuecomment-3716905486 by @Aternus

> [@ozh](https://github.com/ozh) [@LeoColomb](https://github.com/LeoColomb) I've updated the repo with the latest progress so you can see the direction I'm going with, lmk what you think 🙏
> 
> [Aternus/yourls-responsive-ui](https://github.com/Aternus/yourls-responsive-ui?rgh-link-date=2026-01-07T00%3A56%3A53.000Z)
> 
> If you want to try it out you'll need to build the assets by running `npm run build` from the project root (after copying/cloning the project).

punkajk:
I opened the issue #2693 majorly because of the outdated, non-responsive admin area of Yourls...on mobile devices. I expected the changes using CSS only. However, in my opinion, here's what could be the most appropriate, logical, and simplest way to proceed considering the use of Yourls on mobile:
1. Include basic responsive design element in YOURLS source stylesheet.
2. The UI redesign can be done using Themes (in plugins), similar to Sleeky.
3. Using manifest.json or something to make the entire YOURLS instance installable on mobile devices as a PWA.
_Sleeky is okay but not perfect._
Node.js is not okay. Majority of the YOURLS installations are on shared hosting environment, which only support PHP. Please give a feedback on this if you think this is the way forward.
I strongly believe complete responsive design is what we need and then making it installable as PWA would benefit mobile users.

LeoColomb:
I'm able to compile and get it live using the Docker environment.
All fine on that part.

I was not expecting to get an actual implementation straight away 😅
Are the technical choices being made to speed up the first draft, or are they felt as definitive?
Asking this as I don't see the need for SASS, and skipping this would help get rid of Node.js use (at least for now) and reduce @ozh pain 😁

The Docker implementation feels quite overcomplicated, but this is secondary.

Now the visual/design: it's a good first iteration for sure, and it is responsive.
Still more to be done, but I guess that's obvious.

[...]

PR #4053, opened 2026-01-25
Responsive Admin Interface Redesign

dineshingale:
This PR adds responsive design capabilities to the YOURLS admin interface, addressing issue #4045. Key changes include adding viewport meta tag and media queries for mobile devices.

### DIFF
--- a/css/infos.css
+++ b/css/infos.css
@@ -1,116 +1,182 @@
 h3 span.label {
-	width:100px;
-	display:inline-block;
+	width: 100px;
+	display: inline-block;
 }
 
 ul.toggle_display {
-	display:none;
-	list-style-type:none;
-	margin-left:0;
-	margin-right:23px;
-	padding:12px 5px 3px;
-	border-bottom:1px solid #C7E7FF;
+	display: none;
+	list-style-type: none;
+	margin-left: 0;
+	margin-right: 23px;
+	padding: 12px 5px 3px;
+	border-bottom: 1px solid #C7E7FF;
 }
+
 ul.toggle_display li {
-	padding:0;
+	padding: 0;
 }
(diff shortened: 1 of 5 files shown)


ozh:
A few things don't work too well

When plugins have admin pages : 
image

Random spacing in the "new URL" form (might be already there but nothing improved at least), but more importantly the "Action" column makes no sense and a weird grey space appears (randomly?) on the right
image

Overall I'm not sure there's a real added value in just reworking the top left menu to make it horizontal. Am I missing something else?

| `master`  | this PR |
| ------------- | ------------- |
| image  | image  |

toineenzo:
Isn't it a better idea to also give the design a complete overhaul? I think it looks really dated. Sleeky should be a good baseline, maybe a good idea to see how it plays out if this gets merged? Or are we gonna run into BC problems then?

toineenzo:
Ah just saw that https://github.com/Aternus/yourls-responsive-ui exists, never knew of it. Let's see what's the best option.

dgw:
> Isn't it a better idea to also give the design a complete overhaul? I think it looks really dated. Sleeky should be a good baseline, maybe a good idea to see how it plays out if this gets merged? Or are we gonna run into BC problems then?

I don't know what you mean by the latter part, but we have notionally wanted to "redesign" YOURLS for something like a decade. Nobody has had the motivation or spare time to do so, and incremental improvements like responsiveness are still better than doing nothing.

And as you pointed out, "more modern" looks are already available as extension code. Having just been driven away from a site I used to love (since 2013) due to a HORRIBLE redesign-for-the-sake-of-redesigning, I'm more biased than ever against face-lifting anything just because it "looks dated".

toineenzo:
@dgw Thanks for clarifying! BC means Backwards Compatible.

Issue #3937, opened 2025-06-17
Localization support for date and time display


infinitail:
### Proposed functionality

In YOURLS, values like "M d, Y H:i" are passed as arguments to the yourls_date_i18n() function. I would like these arguments to be treated as localization targets using the yourls__() function. While it may be possible to handle this through filters, I believe addressing it through localization is more natural.

### Use case

Site users will be able to use date formats that follow their locale.

### Potential solution

Rewrite code like below. No potential impact for the current user.
```diff
-   yourls_date_i18n( yourls_get_datetime_format('M d, Y H:i'), yourls_get_timestamp( $timestamp ))
+   yourls_date_i18n( yourls_get_datetime_format( yourls__('M d, Y H:i') ), yourls_get_timestamp( $timestamp ))
```

ozh:
Indeed. We translate individual words like day and month, but not the whole date format.
![Image](https://github.com/user-attachments/assets/b29fbf95-6111-4518-94bc-9fad7ebc682f)

PR #4054, opened 2026-01-25
Localization support for date and time display

dineshingale:
This PR adds localization support for date and time display in admin interface and info pages, as requested in issue #3937. It wraps the hardcoded date format strings in yourls__() so they can be translated/customized per locale.


ozh (approved):

ozh:
Thanks !

PR #4058, opened 2026-01-30
Improve shunt filters

ozh:
Fixes #4057

Also, update the shunt logic in GeoIP functions and while I'm at it update the geoip tests


LeoColomb (approved):
LGTM

PR #4064, opened 2026-02-15
Filter SQL queries

ozh:
This one is neat.

It introduces the ability to filter SQL query statements, or completely bypass `fetch*` methods.

Practical uses : 
* rewrite all queries to make things compatible with PostgreSQL without hacking core
* bypass SQL queries to store and retrieve stuff from cache
* and many more stuff of course

Postgresql plugin functional : 
image

Cache plugin functional, support APCu, Redis, Memcache, or writing to disk
image


LeoColomb (commented):
Quite a big chunk to review! 😅
Seems pretty fair on the principle.
That said, is it really needed to reimplement all parent class fetch functions?
As they all call `perform()` anyway, wouldn't it be just good enough to override this one to include the filter?

ozh:
Thanks for reviewing !!

Legit question. The answer is that the `fetchXXX` functions return serializable content, while `perform` returns a PDO statement. I needed to identify the fetch method for a cache plugin concept, without using an ugly debug_backtrace() in `perform` to guess the caller (`fetchOne`, `fetchPairs`, etc...)

LeoColomb:
AI answer, right? 😅
Because "ugly `debug_backtrace()`" is actually used through this change set already. So not really logical…

ozh:
Nope, my own words :)

Regarding the other debugbacktrace : I was  relunctant to use it but couldn't find another way. Eventually used it because it's only used when debug is true. 

In perform it would be always active, which I thought would be a bad idea

ozh:
Is there anything, or anything else, that I may clarify to move on with this PR?

LeoColomb (approved):

PR #4066, opened 2026-02-19
Flag password file as user auth from environment variables

LeoColomb:
Closes https://github.com/YOURLS/containers/pull/468

### DIFF
--- a/includes/functions-auth.php
+++ b/includes/functions-auth.php
@@ -708,8 +708,7 @@ function yourls_verify_nonce($action, $nonce = false, $user = false, $return = '
  * @return bool  true if YOURLS_USER and YOURLS_PASSWORD are defined as environment variables
  */
 function yourls_is_user_from_env() {
-    return yourls_apply_filter('is_user_from_env', getenv('YOURLS_USER') && getenv('YOURLS_PASSWORD'));
-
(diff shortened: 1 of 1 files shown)


ozh:
The container doc mentions `YOURLS_PASS`, not `YOURLS_PASSWORD`. Shouldn't it be `getenv('YOURLS_PASS')` here ?

Also: is there a scenario where we'cdd need to white list `YOURLS_USER_FILE` too ?

LeoColomb:
> The container doc mentions `YOURLS_PASS`, not `YOURLS_PASSWORD`. Shouldn't it be `getenv('YOURLS_PASS')` here ?

It's not the most transparent way of doing things, but no:
* `YOURLS_PASSWORD` is used in YOURLS unit tests
* `YOURLS_PASS` is used by the container, but if `YOURLS_USER` and `YOURLS_PASS` are defined, the container init script is moving them to the config file, and encryption is expected.
  * This behavior is maybe unwanted. It has been imported from other typical official image scripts made by Docker Inc.

> Also: is there a scenario where we'cdd need to white list `YOURLS_USER_FILE` too ?

I guess you're right, indeed.

ozh (approved):
OK for me

LeoColomb:
@ozh Thanks for the review. I just made a second pass after merging https://github.com/YOURLS/containers/pull/482.

It should make function behavior more consistent compared to expectations.
I've dropped the `YOURLS_USER` check since this part is not encrypted and not necessary for the password to be defined as an env var.

ozh (approved):

PR #4069, opened 2026-02-23
Notice overlapping logo in admin panel

### LINKED ISSUE #3578: Notice overlaps with header logo in admin panel
### Code of Conduct

### Submission validity

### Self troubleshooting

### Version

1.9.2

### Description

When e.g. (de)activating a plugin, the `div.notice` overlaps with the header YOURLS logo:

image

### Expectation

_No response_

### Reproduction steps

1. Log into admin panel
2. Go to Manage Plugins
3. Click "Activate" or "Deactivate" for one of the plugins available

### Context

A quick fix for this is to add `max-height: 4.5em;` to the `#yourls-logo` style rules.

There might be a better approach, though. The number of items in the menu can vary depending on which plugins are active and which page of the admin panel the user's on.

_Thanks to @Yuuki2012 for pointing this out, and letting me test CSS tweaks on their instance._

og-khushalpatel:
The `h1` element was using a hardcoded `height: 50px` which, combined with its `float: right`, caused `div.notice` to overlap the logo when shown after plugin activation or deactivation. Block elements do not clear floats by default, so the notice would render at the top of the collapsed header rather than below it.

Two changes made to css/style.css:

- Remove `height: 50px` from `h1` so the header height is no longer artificially constrained to a fixed value.

- Add `clear: both` to `.notice` so it always starts below any preceding floated elements, directly preventing the overlap with the logo.

Fixes #3578

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -32,7 +32,7 @@ a, a:link, a:active, a:visited {
 a:hover {
 	text-decoration: underline;
 }
-h1 {height:50px;margin:0;float:right;max-width:500px;}
+h1 {margin:0;float:right;max-width:500px;}
 h1 a {text-align:right;font-size:20px;float:right;}
 h1 a, h1 a:link, h1 a:active, h1 a:visited {color:#2a85b3}
 h1 a:hover{text-decoration:none;}
@@ -287,6 +287,7 @@ a.bookmarklet:hover {
 	margin-left:15%;
 	padding-left:10px;
 	margin-bottom:5px;
+	clear: both;
 }
 
 .jquery-notify-bar {


og-khushalpatel:
Hi @dgw ,  I’m following up on this PR in case it got buried. Please let me know if any changes are needed from my side. Thank you!

ozh (approved):
LGTM 👍

ozh:
(where is the "Update branch" button ? is the PR branch protected on your side @og-khushalpatel or something like this?)

dgw:
(@ozh The checkbox must be deselected. I get a button anyway provided by the RGH extension, and it tries but then throws a permission error.)

image

og-khushalpatel:
> (where is the "Update branch" button ? is the PR branch protected on your side @og-khushalpatel or something like this?)


I guess it was because i might have disabled the "Allow edits by maintainers" checkbox by mistake while raising this PR. Sorry for that, I have now enabled that checkbox.

PR #4071, opened 2026-02-25
Clear deleted short-url's logs which have became orphan-records in DB.

### LINKED ISSUE #3684: Remove corresponding log entries when a link is deleted
### Code of Conduct

### Submission validity

### Summary

When a link is deleted, the corresponding log entries remain in the DB.

### Motivation

Trying to keep the DB as lean as possible.

### Describe the solution

_No response_

og-khushalpatel:
Previously, when user was deleting the short url record from the admin panel, the stats related to the deleted short url were not being deleted. This resulted in orphaned logs in `your_logs` table.

Implementation:

1. I created a function `yourls_delete_stats_for_keyword()` in `includes/functions-info.php` which will delete the logs/stats corresponding to the deleted short link from the database.
2. Then I hooked `yourls_delete_stats_for_keyword()` to `delete_link` action in `includes/functions-kses.php` file.
3. Finally I added a test method to test if the `yourls_delete_link_by_keyword()` clears the logs after deletion.

Fixes #3684

### DIFF
--- a/.github/workflows/update-translations.yml
+++ b/.github/workflows/update-translations.yml
@@ -49,7 +49,7 @@ jobs:
             --keyword=yourls_n_noop:1,2 \
             --keyword=yourls_nx_noop:1,2
 
-      - uses: actions/upload-artifact@v6
+      - uses: actions/upload-artifact@v7
         with:
           name: YOURLS-pot
           path: YOURLS.pot
@@ -66,7 +66,7 @@ jobs:
         with:
           repository: YOURLS/YOURLS.pot
 
-      - uses: actions/download-artifact@v7
+      - uses: actions/download-artifact@v8
         with:
           name: YOURLS-pot
 
--- a/includes/class-mysql.php
+++ b/includes/class-mysql.php
@@ -96,8 +96,8 @@ function yourls_db_connect($context = '') {
  */
 function yourls_get_db($context = '') {
     // Allow plugins to short-circuit the whole function
-    $pre = yourls_apply_filter( 'shunt_get_db', false, $context );
-    if ( false !== $pre ) {
+    $pre = yourls_apply_filter( 'shunt_get_db', yourls_shunt_default(), $context );
+    if ( yourls_shunt_default() !== $pre ) {
         return $pre;
     }
 
--- a/includes/functions-auth.php
+++ b/includes/functions-auth.php
@@ -25,8 +25,8 @@ function yourls_maybe_require_auth() {
  */
 function yourls_is_valid_user() {
     // Allow plugins to short-circuit the whole function
-    $pre = yourls_apply_filter( 'shunt_is_valid_user', null );
-    if ( null !== $pre ) {
+    $pre = yourls_apply_filter( 'shunt_is_valid_user', yourls_shunt_default() );
+    if ( yourls_shunt_default() !== $pre ) {
         return $pre;
     }
 
--- a/includes/functions-geo.php
+++ b/includes/functions-geo.php
@@ -14,10 +14,12 @@
  */
 function yourls_geo_ip_to_countrycode( $ip = '', $default = '' ) {
     // Allow plugins to short-circuit the Geo IP API
-    $location = yourls_apply_filter( 'shunt_geo_ip_to_countrycode', false, $ip, $default ); // at this point $ip can be '', check if your plugin hooks in here
-    if ( false !== $location ) {
+    $location = yourls_apply_filter( 'shunt_geo_ip_to_countrycode', yourls_shunt_default(), $ip, $default ); // at this point $ip can be '', check if your plugin hooks in here
+    if ( yourls_shunt_default() !== $location ) {
         return $location;
     }
+
+    $location = false;
 
     if ( yourls_apply_filter( 'geo_use_cloudflare', true )
          && !empty( $_SERVER[ 'HTTP_CF_IPCOUNTRY' ] ) && preg_match( '/^[A-Z]{2}$/', $_SERVER[ 'HTTP_CF_IPCOUNTRY' ] ) ) {
@@ -74,10 +76,12 @@ function yourls_geo_ip_to_countrycode( $ip = '', $default = '' ) {
  */
 function yourls_geo_countrycode_to_countryname( $code ) {
     // Allow plugins to short-circuit the function
-    $country = yourls_apply_filter( 'shunt_geo_countrycode_to_countryname', false, $code );
-    if ( false !== $country ) {
+    $country = yourls_apply_filter( 'shunt_geo_countrycode_to_countryname', yourls_shunt_default(), $code );
+    if ( yourls_shunt_default() !== $country ) {
         return $country;
     }
+
+    $country = false;
 
     // Weeeeeeeeeeee
     $countries = [
--- a/includes/functions-html.php
+++ b/includes/functions-html.php
@@ -181,8 +181,8 @@ function yourls_html_footer($can_query = true) {
  * @return void
  */
 function yourls_html_addnew( $url = '', $keyword = '' ) {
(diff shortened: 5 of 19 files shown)


ozh:
(i will review next week)

ozh:
Thanks for this PR (and including unit tests is always appreciated a lot :)

The idea sounds sane but in practice I'm not sure it's a good idea. Here are my thoughts : 

- Minor : we already have `yourls_delete_link_by_keyword()` -- not sure if "by keyword" is much better than "for keyword" but for consistency maybe the function here should be `yourls_delete_stats_by_keyword()` (instead of `for_keyword`)
- Minor : we have recently improved the "shunt" filters : `false` needs to be replaced with `yourls_shunt_default()` -- see https://github.com/YOURLS/YOURLS/pull/4058/
- Major : following the same logic, I guess we should also have a `yourls_edit_stats_by_keyword()`, when someone edits a short URL and modifies the keyword

Now the "is it a good idea?" part

This PR is, in fact, a fix for my poor database design when the project started (using the keyword in both YOURLS_URLS and YOURLS_LOG, instead of having a table with keywords, and using an ID referencing that keyword in other tables) (this is still an improvement that is planned sometimes in the future)

My fear is that it would fix a non existent/minor problem for small users, but introduce a blocking bug for heavy users.

* minor problem : I guess 99% of people don't edit or delete their short URLs
* yet blocking : deleting a keyword with a few hundreds of thousands hits would time out the app, mid operation, because the PHP max exec time would be soon reached. Blocking, or even destructing, since an SQL query that times out can have unexpected effects

og-khushalpatel:
> Now the "is it a good idea?" part

Thanks a lot for the detailed feedback!

- About the potential performance issue you mentioned: you're right that deleting stats for keywords with a very large number of hits could become expensive. 
- The current implementation already performs the deletion using a single SQL statement (`DELETE FROM YOURLS_LOG WHERE shorturl = :keyword`) through the DB layer. This avoids iterating through log entries in PHP and lets the database handle the operation internally, which should scale much better even when a keyword has a large number of log rows. 
- However, I understand that single SQL Delete query can lock large portions of the table, block inserts, create long transactions, etc. 
- For that we can make use of Batched Deletes to avoid all the mentioned concerns but we will have to fire a few extra queries instead of a single one. This trade off is gonna be there so we can find a sweet spot for the batch size that can take care of the performance concern. But I'd be happy to adjust the approach if you think another strategy would be safer.

Finally, would you prefer that I update the PR with the minor/major adjustments suggested by you, or should we first decide whether this feature is going in the right direction considering the future performance concerns?

ozh:
I've been thinking about this over the last few days.

Poor DB design or not, that doesn't change the problem - deleting hundreds of thousands of rows isn't compatible with most (if not all) setups.

Basically I see 3 solutions:
1) Delete content in batches - something like 5,000 by 5,000, with an client ajax loop. Super long, would require user to leave window open till the operation is complete.
2) We could make a "soft delete" slightly over convoluted approach : flag in the option table the keyword to be deleted in logs, and then on each subsequent load, check if there is something to delete and if so delete like 500 rows. Transparent to the admin but would add load on each redirection, again not ideal for heavy installs, and without any view on progress.
3) Just letting as is, dead data.

My bet is that a fair share of people don't often rename or delete URL, and a fair share of that fair share doesn't care (and most likely doesn't know) if their DB has a few MB too much or not. I think we're aiming at fixing a problem most people just don't have.

We could complement the 3rd approach with a plugin for those interested in optimizing and pruning, for users who often delete short URLs, that would use the first approach.

dgw:
A maintenance script is the approach I usually see for this kind of thing in other projects. Something like `php artisan db:cleanup`

YOURLS obviously isn't using Laravel, but this has been a pattern going way back. When I used to run a (personal) MediaWiki, it came with maintenance scripts to clean up various stuff from the CLI.

I kind of like option 2, the soft-delete. If YOURLS finished sending the redirection headers (or the custom/admin page) first, doing a little cleanup before terminating the request wouldn't hurt. Maybe even just 50-100 rows at a time.

But a maintenance script is cleaner…

og-khushalpatel:
@ozh I believe the idea proposed by @dgw about maintenance script sounds cleaner and manageable.

ozh:
Then I think the best is a plugin -- web based for people who have no shell access or do not know how to use -- and a cleanup script

og-khushalpatel:
@ozh @dgw  That sounds like a good approach.

I’m relatively new to open source contributions and haven’t created a YOURLS plugin before, but I’d definitely be interested in implementing this as a plugin and a cleanup script as suggested.

If you could point me to any examples or guidelines for building such a plugin in the YOURLS ecosystem, I’d be happy to work on it and share the result here.

Thanks again for the helpful feedback!

ozh:
Here is how I would do it

A `plugin.php` with the minimum needed - basically just `require` files `functions.php` and `admin.php`

In `functions.php` there would be all funcs needed: to check how many orphan entries exist in the DB, to delete them, maybe to calculate the size of saved space, etc

`admin.php` would handle the admin interface (using function `yourls_register_plugin_page()`, see for instance this plugin https://github.com/ozh/yourls-change-password/) where the user would maybe define the batch size, see how many orphan rows there are, click "Delete orphan rows" and watch a counter decreased.   
This file would also print out the few lines of JS needed to delete in batch in an ajax loop. No need to reinvent the powder: core file `admin/admin-ajax.php` (its URL is `yourls_admin_url('admin-ajax.php')`) allows custom actions (`yourls_do_action( 'yourls_ajax_'.$action );`)

And finally, a `cli.php` that would use functions from `functions.php` and designed to be executed from the CLI directly. Something like :  
```
$ php cli.php --check
Found 13 outdated keywords in 1337 rows (occupying 31337 Mb)
$ php cli.php --delete
Deleted 1337 rows. 31337 Mb saved!
$
```

cbrocas:
Just a feedback from a potential user of Yourls (currently evaluating FLOSS URL shorteners) in a setup that would have potentially several millions (sometimes tens of) of hits per short URLs.

Not having a way to cleanup an always growing table is a major concern in production. It will end badly one day or another.

Having a way to cleanup in a batch manner entries in yourls_logs table corresponding to deleted links would be a good way to handle this situation IMHO: 
- people with small / not heavily used setup (with no CLI access) would simply ignore it.
- people with high load setup would have a solution to try to handle the growth of this table.

og-khushalpatel:
> Here is how I would do it
> 
> A `plugin.php` with the minimum needed - basically just `require` files `functions.php` and `admin.php`
> 
> In `functions.php` there would be all funcs needed: to check how many orphan entries exist in the DB, to delete them, maybe to calculate the size of saved space, etc
> 
> `admin.php` would handle the admin interface (using function `yourls_register_plugin_page()`, see for instance this plugin https://github.com/ozh/yourls-change-password/) where the user would maybe define the batch size, see how many orphan rows there are, click "Delete orphan rows" and watch a counter decreased. This file would also print out the few lines of JS needed to delete in batch in an ajax loop. No need to reinvent the powder: core file `admin/admin-ajax.php` (its URL is `yourls_admin_url('admin-ajax.php')`) allows custom actions (`yourls_do_action( 'yourls_ajax_'.$action );`)
> 
> And finally, a `cli.php` that would use functions from `functions.php` and designed to be executed from the CLI directly. Something like :
> 
> ```
> $ php cli.php --check
> Found 13 outdated keywords in 1337 rows (occupying 31337 Mb)
> $ php cli.php --delete
> Deleted 1337 rows. 31337 Mb saved!
> $
> ```

I am really grateful for the explanation provided by you. Recently I am facing some issues in my laptop as a result not able to contribute. I promise to get on it as soon as my laptop gets repaired:)

ozh:
Closing here because I understand this is (maybe) going onto another direction (plugin + CLI), but feel free to comment on if I'm wrong

PR #4088, opened 2026-04-11
Remove current Angel Sponsors section

ozh:
Text & tags left as comment to be used next time


dgw (approved):
Unrelated to this specific change, is it actually guaranteed that you can just drop an image into some comment box on GH and use the resulting attachment URL, indefinitely?

(Likely a moot question, since the eventual plan is automating these updates.)

ozh:
> Unrelated to this specific change, is it actually guaranteed that you can just drop an image into some comment box on GH and use the resulting attachment URL, indefinitely?

Indefinitely, dunno, but [at least 12 years](https://github.com/ozh/plugin-unit-test-demo) it seems :)

dgw:
That's pretty indefinite 😅\
Good enough for me

PR #4092, opened 2026-04-20
feat(devex): add Docker setup for local development

Aternus:
@ozh I've started backporting the key pieces to facilitate responsive UI with theme and pluggable architecture.

Reference: https://github.com/Aternus/yourls-responsive-ui

Next will be creating a PHP CodeSniffer ruleset and aligning the current codebase so PR changesets are minimal.

## Summary

- Add Docker Compose setup for local YOURLS development with MySQL, Apache/PHP, BrowserSync (live reload), and phpMyAdmin
- Include Xdebug configuration for debugging support out of the box
- Add CONTRIBUTING.md with setup instructions for the Docker-based development environment
- Update .gitignore with comprehensive OS, editor, and build tool exclusions

## Details

The Docker setup includes:
- **web**: PHP 8.1 + Apache with SSL, Xdebug, and common PHP extensions
- **db**: MySQL 8.0
- **browsersync**: Node-based live reload proxy
- **phpmyadmin**: Database management UI on port 8080

Developers can get started with `docker compose up --build` after generating local SSL certs via `mkcert`.

## Test plan

- [ ] Run `docker compose up --build` and verify all services start
- [ ] Access YOURLS admin at `https://localhost/admin` (login: root/root)
- [ ] Verify phpMyAdmin is accessible at `http://localhost:8080`
- [ ] Verify BrowserSync live reload works when editing files
- [ ] Confirm Xdebug connects to IDE on port 9003

PR #4093, opened 2026-04-20
Fix PHP 8.1 compatibility

Aternus:
Following #4092 

This PR ensures the vendor dependecies are compatible with PHP8.1 (project's minimum supported PHP version) **while also fixing related bugs**.

There is a large number of changes b/c the vendor folder.

## Summary

- **Fixed PHP 8.1 compatibility**: downgraded `aura/sql` from v6 to v5 (compatible with PHP 8.1), upgraded `geoip2/geoip2` to v3
- **Fixed bug in `functions-auth.php`**: changed loose comparison (`==`) to strict comparison (`===`) when checking `file_get_contents()` return value, preventing false negatives on empty files
- **Fixed auth tests**: `test_hash_passwords_now_unwritable` now uses `tempnam()` instead of hardcoded path; `test_hash_passwords_now_empty` expects correct error message after the strict comparison fix
- **Improved CI config**: test database credentials are now passed via environment variables instead of a copied config file

## Test plan

- [ ] CI passes on PHP 8.1, 8.2, 8.3, and 8.4
- [ ] Auth tests pass with the strict comparison fix
- [ ] Docker setup works for local development (`docker compose up`)
- [ ] Existing functionality is not affected by dependency changes


ozh:
A typical PR that won't be reviewed, sorry : 

- fixes a few things, some might be good idea, dunno, because
- "fixes" PHP 8.1 which [works fine](https://github.com/YOURLS/YOURLS/actions/runs/24671460985/job/72143146244)
- in a gigantic PR of 103 files changed +3,290 -2,306
- which adds tons of unneeded and unrelated stuff (like docker config)
- and makes useless rewrites like one liners when things are on 3 lines.

Bottom line : PR generated by AI tools and submitted as is are a waste of time for everybody.

I'd rather have 10 PR fixing or adding 10 stuff than 1 PR like this.

PR #4107, opened 2026-05-20
Prevent XSS in stat pages through referrers

ozh:
Fix potential XSS as described [here](https://github.com/YOURLS/YOURLS/security/advisories/GHSA-5h77-88j3-r659) 

* JS data used to generate the graph is handled with `yourls_esc_js()` as it should have been
* A few more sanitization while I was there, with `yourls_get_favicon_url()` and `yourls_get_domain()` to output better HTML
* Minor coding style update limited to the functions I've updated (return type, etc)

### DIFF
--- a/includes/functions-infos.php
+++ b/includes/functions-infos.php
@@ -38,8 +38,8 @@ function yourls_stats_countries_map($countries, $id = null) {
  *
  * @param array $data  Array of 'data' => 'value'
  * @param int $limit   Optional limit list of X first countries
- * @param $size        Optional size of the image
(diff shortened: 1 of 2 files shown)


ozh:
Unit tests don't really care about this patch. XSS resolution confirmed manually: graphs are correctly printed, without extra JS code executed.

dgw (approved):
Long-term I think we should add unit tests for this to guard against regression, but I see nothing wrong with the fixes here.

LeoColomb (approved):

PR #4111, opened 2026-05-27
Misc security hardenings

ozh:
Misc security improvement, per our private discussion.

Probable plan : 

* [x] phase out md5 (**F1**)
* [x] replace `===` with `hash_equals()` where appropriate (**F2**)
* [x] default to REMOTE_ADDR only when no trusted proxy is configured (**F3**).
* [x] increase API signature length (**F4**)
* [ ] ~~remote title fetching (**F5**)~~
* [x] decouple cookie and api sig (**F6**)
* [x] escape HTML in `yourls_notice_box()` (**F7**)

### DIFF
--- a/includes/functions-auth.php
+++ b/includes/functions-auth.php
@@ -357,12 +357,15 @@ function yourls_check_auth_cookie() {
  * Since 1.7.7 we allow a `hash` parameter and an arbitrary hashed signature, hashed
  * with the `hash` function. Examples :
  *   http://sho.rt/yourls-api.php?timestamp=<timestamp>&signature=<sha512 hash>&hash=sha512&action=...
- *   http://sho.rt/yourls-api.php?timestamp=<timestamp>&signature=<crc32 hash>&hash=crc32&action=...
+ * Since 1.10.5, the hash must be one of: sha256, sha384, or sha512, unless explicitly allowed by a plugin via the
+ * `allowed_hash_algos` filter.
+ *
+ * @see https://yourls.org/docs/guide/advanced/passwordless-api
  *
  * @since 1.4.1
  * @return bool False if signature or timestamp missing or invalid, true if valid
  */
-function yourls_check_signature_timestamp() {
+function yourls_check_signature_timestamp(): bool {
     if(   !isset( $_REQUEST['signature'] ) OR empty( $_REQUEST['signature'] )
        OR !isset( $_REQUEST['timestamp'] ) OR empty( $_REQUEST['timestamp'] )
     ) {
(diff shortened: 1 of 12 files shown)


LeoColomb (dismissed):

ozh:
Phasing out md5 passwords in config.php and helping the user improve: 

image

ozh:
I'll ignore the **F5** case. While the point is valid in theory, it ignores the use case where YOURLS would be deployed entirely over a private network, like a corporate intranet. Happy to expand my thoughts in private if needed :)

ozh:
Ready for review :tada: 

Turned out a bit bigger of a patch than anticipated, so I tried to keep each commit tied to a specific item (**F-something** as listed in the PR desc)

Everything deprecated is silently updated, or gracefully de-phased and advertised as deprecated, without breaking changes.

Roughly a hundred new assertions added in tests, and several visual & manual inspections where unit tests don't apply (JS behavior in the interface for instance)

includes/functions-auth.php
LeoColomb: Could be replaced with `str_starts_with()`

includes/functions.php
```diff
@@ -17,29 +17,92 @@ function yourls_make_regexp_pattern( $string ) {
...
         }
     }
 
-    // headers can contain multiple IPs (X-Forwarded-For = client, proxy1, proxy2). Take first one.
-    if ( strpos( $ip, ',' ) !== false )
-        $ip = substr( $ip, 0, strpos( $ip, ',' ) );
+    // If there are multiple IPs (e.g. in HTTP_X_FORWARDED_FOR), take the first one
+    if (str_contains($ip, ',')) {
+        $ip = substr($ip, 0, strpos($ip, ','));
+    }
```
LeoColomb:
```php
$ip = explode(',', $ip, 1)[0]
```

includes/functions-auth.php
```diff
@@ -463,16 +463,29 @@ function yourls_check_signature() {
...
+ * @param false|string $username Username to generate signature for, or false to use current user
  * @return string                 Signature
  */
-function yourls_auth_signature( $username = false ) {
+function yourls_auth_signature(false|string $username = false ): string {
     if( !$username && defined('YOURLS_USER') ) {
         $username = YOURLS_USER;
     }
-    return ( $username ? substr( yourls_salt( $username ), 0, 10 ) : 'Cannot generate auth signature: no username' );
+    $signature = $username ? substr(yourls_salt($username ), 0, yourls_auth_signature_length() ) : 'Cannot generate auth signature: no username';
```
LeoColomb: `substr` => `str_starts_with`

LeoColomb (approved):
It seems fairly OK. I've not read the test files.
I'll be honest, the whole PR is complicated to review with everything all together, even with the commit chuncks. A PR per scope next time, maybe? 😁

Just a couple of inline comments, not critical at all, optional to apply them.
Oh and a commit message to rewrite, `@since` contains a tag `@` which then tags a GitHub user. As per experience, being tagged on a commit is hell! 🥲

includes/functions.php
ozh: Nicer, indeed - but actually wrong, since with a limit of `, 1` it doesn't split ;) Needs a limit of 2, or just no limit. See https://3v4l.org/mDloY

includes/functions-auth.php
ozh: Don't think so : this `substr()` doesn't test the beginning of the string, it extracts the beginning

includes/functions-auth.php
ozh: :+1:

PR #4112, opened 2026-05-30
Update dependencies

ozh:
- Upgrading composer/ca-bundle (1.5.5 => 1.5.12)
  - Upgrading maxmind-db/reader (v1.12.0 => v1.13.1)
  - Upgrading maxmind/web-service-common (v0.10.0 => v0.11.1)
  - Upgrading ozh/bookmarkletgen (1.2.2 => 1.3)
  - Upgrading rmccue/requests (v2.0.15 => v2.0.18)
  - Upgrading symfony/polyfill-intl-idn (v1.31.0 => v1.38.1)
  - Upgrading symfony/polyfill-intl-normalizer (v1.31.0 => v1.38.0)
  - Upgrading symfony/polyfill-mbstring (v1.31.0 => v1.38.1)

(security updates in symfony/polyfill)


ozh:
The butchered version of aura/sql, pretending to require 8.4+, breaks things ?

I don't know how to update composer stuff because of this.

dgw:
The thing with aura/sql was a big part of why I had trouble getting a local dev environment set up properly this last week. I should be able to `composer install` the project and start working, but nope, it needs handholding.

ozh:
So, this "+7,549 -1,503" PR can be summarized with : 

```sh
# don't check local version of PHP
composer config platform-check false
# update everything but aura/sql
composer update --with-all-dependencies  --no-scripts -- ozh/bookmarkletgen rmccue/requests pomo/pomo geoip2/geoip2 jakeasmith/http_build_url symfony/polyfill-mbstring symfony/polyfill-intl-idn spatie/array-to-xml
# clean up unneeded files
./includes/vendor/build-script/yourls-build.sh -d ./includes/vendor/
```

LeoColomb (changes requested):
> The butchered version of aura/sql, pretending to require 8.4+, breaks things ?

Yes, aura/sql requires some specific care, with manual edits (I hate it, but that was the only way last time).

> ```shell
> # don't check local version of PHP
> composer config platform-check false
> ```

When updating dependencies, this feels quite wrong 😅 
It's not checking the local version necessarily, but check the platform compatibility. That's the only way to ensure compatibility with the PHP version announced.
In here, the check is made against this parameter:

https://github.com/YOURLS/YOURLS/blob/f01c08f2201479500838101703518fac6f245876/composer.json#L41-L43

ozh:
Sorry I don't understand what you're suggesting here. Without `composer config platform-check false` builds fail for PHP 8.1 - 8.3

ozh:
I think what we want is:

```json
{
    "require": {
        "php": "^8.1"
    },
    "config": {
        "platform": {
            "php": "8.1.0"
        },
        "platform-check": false
    }
}
```

Min is 8.1, when updating make it like it's 8.1+ to pick dependencies, but don't check PHP version at runtime. @LeoColomb thoughts?

LeoColomb:
I'm on holiday this week, so can't have a deep look right now. 
But no, platform check ensure all dependencies are all compatible between each others, local ones included. We must no disable that one, not even when updating. If it fails, it's for a reason, this should not be overlooked, IMO.

ozh:
The reason it fails is because [vendor/composer/platform_checks.php](https://github.com/YOURLS/YOURLS/pull/4112/changes/717b1776e217013fbaa26af20bad99e0e05f2282#diff-57b53f01efb61f2cdfb4b610fb36d5f0a07c9fe1ead8c1f0ebceaabab3af0487) checks if `PHP_VERSION_ID >= 80400`.

The only thing that `composer config platform-check false` does is remove that check at runtime. Having `config.platform.php` ensures correct versions of deps are selected when updating.

The only thing that `composer config platform-check true` does is add `includes/vendor/composer/platform_check.php`.

My understanding :)

LeoColomb (approved):
Ok, deeply reviewed and fine by me.
I'm not super happy with the platform check value in the `package.json`, but I can live with it.
**BUT** the day we update AuraSQL, this platform check issue must be reviewed again. The platform check is checking 8.4 because of AuraSQL. Last time we updated it, I made manual changes so it is kept compatible with 8.1 (sorry for the lack of documentation, my mistake).

PR #4113, opened 2026-05-30
Remove phpunit6-compat.php

ozh:
Empty file and not referenced anywhere


dgw (approved):
Huh, I wonder what that was there for.

ozh:
Old piece of code to maintain compat between incompatible versions of PHPUnit -- these folks have adopted breaking changes as a lifestyle :)

PR #4114, opened 2026-05-31
Test DB versions

ozh:
In the official doc: 

> Database: YOURLS supports either MySQL (version 5.0 or greater) or MariaDB (version 10.0 or greater)

I suspect we absolutely don't support this :-)

Little matrix test to see what can be checked. This isn't intended to be triggered on each commit/PR

### DIFF
--- a/.github/workflows/db-version.yml
+++ b/.github/workflows/db-version.yml
@@ -3,7 +3,7 @@ name: DB matrix
 on:
   workflow_dispatch:
   push:
-    branches: [db-matrix]
+    branches: [db-matrix-test]
     
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
--- a/.github/workflows/db-version.yml
+++ b/.github/workflows/db-version.yml
@@ -1,9 +1,7 @@
-name: DB matrix
+name: DB matrix compat check
 
 on:
   workflow_dispatch:
-  push:
-    branches: [db-matrix-test]
     
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
@@ -25,12 +23,9 @@ jobs:
           - { image: 'mysql:5.7',     php: '8.5', health: 'mysqladmin ping' }
           - { image: 'mysql:8.0',     php: '8.5', health: 'mysqladmin ping' }
           - { image: 'mysql:8.4',     php: '8.5', health: 'mysqladmin ping' }
-          # 9.0 is an innovation release, may have been superseded by 9.7
           - { image: 'mysql:9.0',     php: '8.5', health: 'mysqladmin ping' }
           - { image: 'mysql:9.7',     php: '8.5', health: 'mysqladmin ping' }
           # --- MariaDB (official `mariadb` image) ---
-          # 5.5 is very old, old manifest may still pull
-          - { image: 'mariadb:5.5',   php: '8.5', health: 'healthcheck.sh --connect --innodb_initialized' }
           - { image: 'mariadb:10.11', php: '8.5', health: 'healthcheck.sh --connect --innodb_initialized' }
           - { image: 'mariadb:11.4',  php: '8.5', health: 'healthcheck.sh --connect --innodb_initialized' }
           - { image: 'mariadb:11.8',  php: '8.5', health: 'healthcheck.sh --connect --innodb_initialized' }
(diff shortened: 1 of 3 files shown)


ozh:
I suggest we require a min of MySQL 5.5 (no image to test for 5.0 and 5.1) and Maria 10 then.

ozh:
Anything useful to be done with this workflow ? Give a cron job once every semester ?

dgw:
I think it's fine to have a manual trigger (the `workflow_dispatch:` one you already included).

ozh:
(Also removed failing matrix entry)

.github/workflows/db-version.yml
```diff
@@ -0,0 +1,71 @@
...
+          - { image: 'mysql:5.7',     php: '8.5', health: 'mysqladmin ping' }
+          - { image: 'mysql:8.0',     php: '8.5', health: 'mysqladmin ping' }
+          - { image: 'mysql:8.4',     php: '8.5', health: 'mysqladmin ping' }
+          - { image: 'mysql:9.0',     php: '8.5', health: 'mysqladmin ping' }
+          - { image: 'mysql:9.7',     php: '8.5', health: 'mysqladmin ping' }
+          # --- MariaDB (official `mariadb` image) ---
+          - { image: 'mariadb:10.11', php: '8.5', health: 'healthcheck.sh --connect --innodb_initialized' }
+          - { image: 'mariadb:11.4',  php: '8.5', health: 'healthcheck.sh --connect --innodb_initialized' }
+          - { image: 'mariadb:11.8',  php: '8.5', health: 'healthcheck.sh --connect --innodb_initialized' }
+          - { image: 'mariadb:12.2',  php: '8.5', health: 'healthcheck.sh --connect --innodb_initialized' }
```
LeoColomb: If the PHP version is the same for all, I'd recommend to separate it from the matrix.
And actually, if the PHP version is not a consideration (which raises its question by itself), I'd just drop the variable completely and let `shivammathur/setup-php` choose the best option (latest stable or the one defined in `composer.json`).

.github/workflows/db-version.yml
```diff
@@ -0,0 +1,71 @@
...
+          MYSQL_ALLOW_EMPTY_PASSWORD: false
+          MARIADB_MYSQL_LOCALHOST_USER: 1
+          MARIADB_MYSQL_LOCALHOST_GRANTS: USAGE
+        options: >-
+          --health-cmd="${{ matrix.health }}"
+          --health-interval=10s
+          --health-timeout=5s
+          --health-retries=10
+
+    steps:
```
LeoColomb: All the steps are copy-paste from `ci.yaml`, right? If so, a composite workflow would probably be better.

.github/workflows/db-version.yml
```diff
@@ -0,0 +1,71 @@
...
+      matrix:
+        include:
+          # --- MySQL (official `mysql` image) ---
+          # 5.5 removed from docker-library (EOL), old manifest may still pull
+          - { image: 'mysql:5.5',     php: '8.5', health: 'mysqladmin ping' }
+          - { image: 'mysql:5.7',     php: '8.5', health: 'mysqladmin ping' }
+          - { image: 'mysql:8.0',     php: '8.5', health: 'mysqladmin ping' }
+          - { image: 'mysql:8.4',     php: '8.5', health: 'mysqladmin ping' }
+          - { image: 'mysql:9.0',     php: '8.5', health: 'mysqladmin ping' }
+          - { image: 'mysql:9.7',     php: '8.5', health: 'mysqladmin ping' }
```
LeoColomb: How are the versions chosen? Why testing 9.7 and 9.0? And why not 9.1?

LeoColomb (changes requested):

.github/workflows/db-version.yml
ozh: PHP versions are tested in ci.yml so the idea was to use any, latest stable, and just focus on DB engine versions. I'll look at delegating the choice to setup-php, good idea :+1:

.github/workflows/db-version.yml
ozh: I'll look into this :+1:

.github/workflows/db-version.yml
ozh: First and (current) last of the `9.x` branch, same for `8.x`. I thought it would be overkill to test every point release

ozh:
Attempt with reusable workflows fails. I'd rather live with a standalone ci.yml and a more or less duplicate that will be used once a year :grin:

PR #4116, opened 2026-05-31
New action to allow custom redirect logging

### LINKED ISSUE #3990: Logging the `Host` header value
### Proposed functionality

Logging the value of the `Host` header (so whatever domain the user clicked on) in the `yourls_log` table. Eventually it could be added to the UI, but I don't mind looking in the datatabase for this.

### Use case

I've got multiple domains (say, foo.example.org and bar.example.org), and want to know which one does better. So I share both, and by checking which one is more popular I know which one to use.

### Potential solution

_No response_

ozh:
Trivial stuff.

Closes #3990

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -535,6 +535,9 @@ function yourls_log_redirect( $keyword ) {
         'location' => yourls_geo_ip_to_countrycode($ip),
     ];
 
+    // Action to allow plugins to log the redirect in their own way. See #3990
+    yourls_do_action( 'log_redirect', $binds, yourls_get_yourls_site() );
+
     // Try and log. An error probably means a concurrency problem : just skip the logging
     try {
         $result = yourls_get_db('write-log_redirect')->fetchAffected("INSERT INTO `$table` (click_time, shorturl, referrer, user_agent, ip_address, country_code) VALUES (:now, :keyword, :referrer, :ua, :ip, :location)", $binds );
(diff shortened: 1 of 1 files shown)


includes/functions.php
```diff
@@ -535,6 +535,9 @@ function yourls_log_redirect( $keyword ) {
         'location' => yourls_geo_ip_to_countrycode($ip),
     ];
 
+    // Action to allow plugins to log the redirect in their own way. See #3990
+    yourls_do_action( 'log_redirect', $binds, yourls_get_yourls_site() );
```
dgw: I'm not sure this fulfills the original feature request, which was to log the `Host` header value. Can't multiple domains point to the same YOURLS instance with a single config and only one of those domains mentioned in `YOURLS_SITE`?

ozh: Actually passing other arguments than `$binds` is useless : if needed, they can be computed in the function hooked to that action (the host header value, for instance).

includes/functions.php
```diff
@@ -526,7 +526,7 @@ function yourls_log_redirect( $keyword ) {
 
     $table = YOURLS_DB_TABLE_LOG;
     $ip = yourls_get_IP();
-    $binds = [
+     = [
```
dgw: Removing the variable name looks like a typo.

```suggestion
    $binds = [
```

dgw (approved):
Now it truly is "Trivial stuff" :)

ozh:
(thanks, truly a typo obviously)

PR #4119, opened 2026-06-03
Implement cookie prefixes

### LINKED ISSUE #2785: Enhance cookies with prefixes ?
As of writing, Firefox and Chrome support "cookie prefixes". Investigate this.

> The `__Secure-` prefix makes a cookie accessible from HTTPS sites only. A HTTP site can not read or update a cookie if the name starts with `__Secure-`. This protects against the attack we earlier described, where an attacker uses a forged insecure site to overwrite a secure cookie.
>
> The `__Host-` prefix does the same as the `__Secure-` prefix and more. A `__Host-`-prefixed cookie is only accessible by the same domain it is set on. This means that a subdomain can no longer overwrite the cookie value.

References:
* https://www.sjoerdlangkemper.nl/2017/02/09/cookie-prefixes/
* https://googlechrome.github.io/samples/cookie-prefixes/

ozh:
Idea surfaced 6 years ago, and [browser support](https://caniuse.com/mdn-http_headers_set-cookie_host_secure_prefixes) is mature now.

Add `__Host-` / `__Secure-` cookie name prefixes to the auth cookie for browser-side hardening on HTTPS installs.

Closes #2785

### DIFF
--- a/includes/functions-auth.php
+++ b/includes/functions-auth.php
@@ -451,7 +451,7 @@ function yourls_check_timestamp( $time ) {
  * @param string $user  User login, or empty string to delete cookie
  * @return void
  */
-function yourls_store_cookie( $user = '' ) {
+function yourls_store_cookie(string $user = '' ): void {
 
     // No user will delete the cookie with a cookie time from the past
     if( !$user ) {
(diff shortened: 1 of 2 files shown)


dgw (approved):
I know a change is straightforward when unit tests far outweigh the actual patch 😁

PR #4122, opened 2026-06-08
Fix early yourls_esc_html()

ozh:
A possible case I didn't take into account in https://github.com/YOURLS/YOURLS/pull/4111 when adding several `yourls_esc_html()` and other sanitizations :

The `yourls_esc_` functions rely on `$yourls_allowedentitynames` being populated, which normally occurs on `plugins_loaded` event: https://github.com/YOURLS/YOURLS/blob/15a9ad52e39c98c53512d65b3f202679224cd091/includes/functions-kses.php#L46

But when the app `yourls_die()` before loading plugins, for example on a bad `config.php` with a wrong DB name, `$yourls_allowedentitynames` is empty, triggering a fatal error down the chain of func calls (`Fatal error: Uncaught TypeError: in_array(): Argument #2 ($haystack) must be of type array, null given in /var/www/html/ozh.in/includes/functions-kses.php on line 607`)

This fixes this issue + adds a relevant unit test

### DIFF
--- a/includes/functions-html.php
+++ b/includes/functions-html.php
@@ -529,7 +529,10 @@ function yourls_die( string $message = '', string $title = '', int $header_code 
     $message = yourls_apply_filter( 'die_message', "<p>$message</p>" );
     $allowed = [
         'strong' => [],
-        'h2'   => [],
+        'em'     => [],
+        'tt'     => [],
+        'code'   => [],
+        'h2'     => [],
         'br'     => [],
         'p'      => ['class' => true],
     ];
--- a/includes/functions-html.php
+++ b/includes/functions-html.php
@@ -524,8 +524,16 @@ function yourls_die( string $message = '', string $title = '', int $header_code 
         yourls_html_head();
         yourls_html_logo();
     }
-    echo yourls_esc_html( yourls_apply_filter( 'die_title', "<h2>$title</h2>" ) );
-    echo yourls_esc_html( yourls_apply_filter( 'die_message', "<p>$message</p>" ) );
+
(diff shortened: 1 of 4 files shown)


includes/functions-html.php
```diff
@@ -524,8 +524,16 @@ function yourls_die( string $message = '', string $title = '', int $header_code
         yourls_html_head();
         yourls_html_logo();
     }
-    echo yourls_esc_html( yourls_apply_filter( 'die_title', "<h2>$title</h2>" ) );
-    echo yourls_esc_html( yourls_apply_filter( 'die_message', "<p>$message</p>" ) );
+
```
ozh, line 527: This fixes the yourls_die() message being wrongly escaped : now allowing some HTML

includes/functions-html.php
```diff
@@ -524,8 +524,16 @@ function yourls_die( string $message = '', string $title = '', int $header_code
...
+    $title = yourls_apply_filter( 'die_title', "<h2>$title</h2>" );
+    $message = yourls_apply_filter( 'die_message', "<p>$message</p>" );
+    $allowed = [
+        'strong' => [],
+        'h2'   => [],
+        'br'     => [],
+        'p'      => ['class' => true],
+    ];
```
dgw, line 538: I suggest a slightly more permissive set of tags:

```suggestion
    $allowed = [
        'strong' => [],
        'em'     => [],
        'tt'     => [],
        'code'   => [],
        'h2'     => [],
        'br'     => [],
        'p'      => ['class' => true],
    ];
```

dgw (approved):
:shipit:

Issue #2772, opened 2020-10-18
Test suite should cover various config cases


ozh:
Issue #2766 arose because the unit tests dictate that `YOURLS_UNIQUE_URLS` is `true` among other stuff, see 
https://github.com/YOURLS/YOURLS/blob/1.7.9/tests/data/config/yourls-tests-config-travis.php#L39

We should have a test suite running and passing under various scenario : `YOURLS_UNIQUE_URLS` true or false, `YOURLS_PRIVATE` true or false, `YOURLS_URL_CONVERT` 32 or 62 or whatever, and so on

ozh:
To do so : make sure each const has a filtered getter (eg `YOURLS_PRIVATE` -> `yourls_is_private()`) and add relevant tests in each group.

See https://github.com/YOURLS/YOURLS/issues/4068

PR #4124, opened 2026-06-09
Make tests independent of constants ; remove some const and define getters

ozh:
Big one, you're going to hate me sorry :) but mostly tests (and a lot of them)

This one tackles #4068 and #2772

The initial idea in #2772 was that tests should run the same no matter the value of any constant (`YOURLS_DEBUG`, `YOURLS_SOMETHING`).  
Option A: run a hellish matrix of tests with all cases covered.  
Option B : define getters for every constant and run test with relevant (filtered) value of that constant

Option B led to defining filterable getters for mostly every const (except structural ones such as `YOURLS_ABSPATH` or `YOURLS_COOKIEKEY`).

The other idea was flawed : in #4068 I wanted to explicitely define all const with a default value. Instead I went the other way: rely on getters and REMOVE some consts.

In short : less consts, consts are only used in one place (the getter), and more tests.

Big patch, but everything was very iterative so it was impossible to split in smaller PRs with days between approvals.

Oh, and totally not a goal in itself but, as I got closer, well, it became a little goal in itself : 
image
1000 tests, baby :wink:


ozh:
Basically, the only non structural const I've left is `YOURLS_DEBUG`. Not sure why but I was more reluctant to remove this one :)

includes/Config/Config.php
```diff
@@ -177,30 +177,6 @@ public function define_core_constants(): void {
...
-        if (!defined( 'YOURLS_NONCE_LIFE' ))
-            define( 'YOURLS_NONCE_LIFE', 43200 ); // 3600 * 12
-
-        // if set to true, disable stat logging (no use for it, too busy servers, ...)
-        if (!defined( 'YOURLS_NOSTATS' ))
-            define( 'YOURLS_NOSTATS', false );
-
-        // if set to true, force https:// in the admin area
-        if (!defined( 'YOURLS_ADMIN_SSL' ))
-            define( 'YOURLS_ADMIN_SSL', false );
```
dgw, line 202: Just using the existing getter, I take it?

https://github.com/YOURLS/YOURLS/blob/d0ad0b786874bbb04882b52d34657aa6101e31ab/includes/functions.php#L913-L920

tests/tests/auth/AuthTest.php
```diff
@@ -82,12 +82,8 @@ public function test_has_phpass() {
...
     {
         yield array( rand_str() );
-        yield array( 'lol .\+*?[^]$(){}=!<>|:-/' . "'" . '"' );
-        yield array( 'أنا أحب النقانق' );
-        yield array( '"double quotes"' );
-        yield array( "'single quotes'" );
-        yield array( '@$*' );
-        yield array( 'أنا أحب النقانق' );
+        yield array( '@$*lol .\+*?[^]$(){}=!<>|:-/' . "'" . '"' ); // funky chars
+        yield array( 'أستمع إلى موسيقى الميتال 🤘️' );              // RTL and UTF-8
```
dgw, line 86: I don't get this one. Why remove the extra cases?

dgw (commented):
I ran out of time to look over _all_ the `tests/` changes, but at least I have looked at all of the `includes/` diffs.

includes/Config/Config.php
ozh: Exactly

tests/tests/auth/AuthTest.php
ozh: Because this test is slow, several seconds (by design, hashing is CPU heavy). So I figured we could just regroup tests, and save hours of CPU in the long run

LeoColomb (commented):
Same as @dgw, non-tests change LGTM; tests seem fine but no bandwidth for a complete review.

PR #4133, opened 2026-07-08
Return HTTP 409 Conflict (not 400) when a duplicate long URL is rejected

### LINKED ISSUE #3547: Short urls http 400 with YOURLS_UNIQUE_URLS=true
### Code of Conduct

### Submission validity

### Self troubleshooting

### Version

1.9.2

### Description

When using the short url api /yourls-api.php?action=shorturl&url=... with YOURLS_UNIQUE_URLS=true the API request works fine on the first time. On the second and consecutive requests for the same url the short url is returned ok but the http response code is 400. 

I tried also with version 1.8.2 and with that it works as expected.

Could be related to the changes in [#3355](https://github.com/YOURLS/YOURLS/pull/3355)

### Expectation

HTTP reponse code 200

### Reproduction steps

_No response_

### Context

_No response_

TowyTowy:
## Pre-requisites

- [x] I have read the [Contributing Guidelines](https://github.com/YOURLS/.github/blob/main/CONTRIBUTING.md)
- [x] I am using AI and have read the [AI Policy](https://github.com/YOURLS/.github/blob/main/AI_POLICY.md)

## Change Description

When `YOURLS_UNIQUE_URLS` is enabled and a long URL that already exists is submitted again, the API returns the existing short URL in the body but sets HTTP status 400. Strict HTTP clients such as n8n or curl with `--fail` treat any 4xx as a failure and never read the body, so they cannot reuse the short URL they were just given. That is the problem reported in #3547.

The change is one line: this specific case now returns 409 Conflict instead of 400, which is what @dgw suggested in the issue thread. Genuinely malformed input (`error:nourl`, `error:noloop`) still returns 400. 409 is already a supported status in `yourls_get_HTTP_status()`. A regression test is added to `DuplicateLongURLTest`; it fails on the current code and passes with the change.

AI disclosure, as the policy asks: I used Claude (Anthropic). It investigated the issue, wrote the patch and the test, ran the test suite against MariaDB in Docker, and helped draft this description. I run YOURLS on my own domain and use its API from my own tooling. I have reviewed the change and understand it: one status code constant changes from 400 to 409 in the duplicate URL branch of `yourls_add_new_link()`, plus the test.

Fixes #3547

### DIFF
--- a/includes/functions-shorturls.php
+++ b/includes/functions-shorturls.php
@@ -85,7 +85,7 @@ function yourls_add_new_link( $url, $keyword = '', $title = '', $row_id = 1 ) {
             yourls_trim_long_string($url), preg_replace('!https?://!', '',  yourls_get_yourls_site()) . '/'. $url_exists->keyword );
         $return['title']    = $url_exists->title;
         $return['shorturl'] = yourls_link($url_exists->keyword);
-        $return['errorCode'] = $return['statusCode'] = '400'; // 400 Bad Request
+        $return['errorCode'] = $return['statusCode'] = '409'; // 409 Conflict: the URL already exists, and the existing short URL is returned
 
         return yourls_apply_filter( 'add_new_link_already_stored_filter', $return, $url, $keyword, $title );
     }
--- a/tests/tests/shorturl/DuplicateLongURLTest.php
+++ b/tests/tests/shorturl/DuplicateLongURLTest.php
@@ -46,4 +46,30 @@ class DuplicateLongURLTest extends PHPUnit\Framework\TestCase {
         $this->assertEquals( 'success', $newurl['status'] );
     }
 
+    /**
+     * When a duplicate long URL is rejected (unique URLs mode), the API must
+     * return HTTP 409 Conflict -- not 400 Bad Request. The request was well
+     * formed and a usable short URL is returned in the response body, so 400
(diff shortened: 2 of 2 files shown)


dgw:
Where is the PR template?

```markdown
## Pre-requisites

- [ ] I have read the [Contributing Guidelines](https://github.com/YOURLS/.github/blob/main/CONTRIBUTING.md)
- [ ] I am using AI and have read the [AI Policy](https://github.com/YOURLS/.github/blob/main/AI_POLICY.md)


## Change Description
<!--- Describe IN YOUR OWN WORDS your changes in detail. Why is this change required? What problem does it solve? --->
```

TowyTowy:
Sorry about that, I skipped the template. The description now follows it, including the AI disclosure per the policy. The change itself is the one line you suggested in the issue, 400 to 409 for the duplicate URL case, plus a regression test that fails on current code and passes with the change.

dgw (approved):
Looks sane to me.

@ozh can you think of any reason not to pull this behavior into core? It's already doable with a plugin (a modified version of https://github.com/YOURLS/YOURLS/issues/3547#issuecomment-2205987214) but I don't see why not to switch from 400 to 409 for everyone.

ozh:
Fine with me as well

LeoColomb (approved):
Looks good to me as well. Merging.

PR #4141, opened 2026-08-03
Restrict remote title fetch to public hosts for anonymous users

ozh:
As discussed in https://github.com/YOURLS/YOURLS/security/advisories/GHSA-gcm8-h82c-vc5p : 

`yourls_get_remote_title()` fetches whatever URL it's given, server side, to read its `<title>`.  
On a public install (`YOURLS_PRIVATE` false, or a public interface on top of a private YOURLS instance), this can be used to probe hosts & ports.
On a private install, this is considered out of scope and none of our business: authenticated users are hopefully trusted, and the admin can drop an arbitrary file to perform any scan.

### How it works

`yourls_get_remote_title()` asks `yourls_restrict_remote_title_fetch()` whether filtering applies. It returns true when constant `YOURLS_USER` is undefined.

When filtering applies, the URL host goes through `yourls_host_is_local()`: an IP literal is validated as is, a host name is resolved first (A and AAAA), and anything that is not a public address, or that doesn't resolve at all, is rejected. On rejection the function returns the URL, which is already what it does when it finds no title, and the link is obviously still created.

Redirects are still followed. Dropping them would break `http -> https`, `example.com -> www.example.com` and valid cases. Instead every hop is checked before it's requested, through a callback on Requests' `requests.before_redirect` hook, to deal with the situation where `attacker.com` would redirect to `10.0.0.1`.

### Allowing all remote title fetches

Would anyone want to always allow the old behavior (for example a public install deliberately running on an intranet) they can opts out with a one liner plugin.

```php
yourls_add_filter( 'restrict_remote_title_fetch', 'yourls_return_false' );
```

**To do**: Ponder whether we should document this on yourls.org/docs

### Things left out of scope from the initial report

- DNS rebinding. Not worth it for a <title> sized vuln. People setting up DNS with 0 TTL will not use YOURLS to scan a host.
- Allow and deny lists. We're not adding more settings, but use filters instead if needed.
- Turning off the automatic title fetch on public installs. Way too strict.

### Tests

Introducing here a new way to run scenarios in a fresh process to test file inclusion and constant definition. I like it :)


ozh:
Ready for review. As always, most of the `+883-27` is tests. Commits should make reviewing easier than the whole 12 changed files diff.

includes/functions-http.php
```diff
@@ -204,6 +204,207 @@ function yourls_send_through_proxy( $url ) {
...
+ * Redirects are still followed: dropping them would break 'http -> https', 'example.com ->
+ * www.example.com', URL shorteners, or any legit 30x redirect. Instead, every hop
+ * is checked before it is requested.
+ *
+ * Meant to be merged into the $options of a single yourls_http_*() call, not to be added to
+ * yourls_http_default_options() -- other requests (core version check, plugins) are not
+ * triggered by an untrusted party.
+ *
+ * @since 1.10.5
+ * @return array  Options to pass to yourls_http_get() & al
```
dgw: Did this line get cut off?

dgw (approved):
Just noticed one odd comment line that ended unexpectedly. Overall, impressed with the new test setup 💪

includes/functions-http.php
ozh: Heh. Professional deformation :) I read a lot and write a few commercial contract documents, where "et al", that we abbreviated as "& al", is used an a lot. It just means "etc...". I'll update, it just looks either pedantic or, as you thought, truncated :)

PR #4145, opened 2026-08-12
SSRF & local IP followup

ozh:
Addressing https://github.com/YOURLS/YOURLS/security/advisories/GHSA-9gg9-rc4w-rj8p

This PR adds checks for IPv6 transition address extraction, oh my, who would have thought the IP address subject would be so deep :rofl: 

A brief wrap up of things hopefully covered in this PR and its prequel #4141 :

|   |  Form     |   What happens when you connect    |                                        Needs                                         |
|-------------|------------------------------------|--------------------------------------------------------------------------------------|-----------------------|
| Notational  | ::ffff:127.0.0.1                   | Nothing IPv6 at all. The kernel emits a plain IPv4 packet.                          | Nothing               |
| Translation | 64:ff9b::7f00:1 (NAT64)            | Real IPv6 packet leaves the host; a gateway rewrites the headers into an IPv4 packet | NAT64 gateway         |
| Tunneling   | 2002:…:: (6to4), 2001:0:… (Teredo) | Real IPv6 packet, wrapped inside an IPv4 packet, sent to the encoded IPv4 endpoint.  | Relay / tunnel client |

I think we're covering an awful lot of unlikely scenarios but at least the topic should be complete now... Except for the case of DNS rebinding (which, for some reason hasn't been reported yet, what are these Claude security scanners doing seriously :) so I added a note explaining why we are NOT covering that case.


dgw (approved):
Couldn't get much more comprehensive if you tried! (Well technically you _could_, but that's addressed by explicitly declaring DNS rebinding out-of-scope.)

Issue #4143, opened 2026-08-09
yourls_get_remote_title() throws TypeError when response has no Content-Type header


bhelmick:
### Self troubleshooting

- [x] I've read the [documentation](https://yourls.org/) and made sure to follow it.
- [x] I've read the [troubleshooting first steps](https://docs.yourls.org/guide/troubleshooting/first-steps.html) and [frequent issues](https://docs.yourls.org/guide/troubleshooting/frequent-issues.html) guides.

### Version

1.10.4

### Description

When shortening a URL whose remote response does not contain a usable Content-Type header, yourls_get_remote_title() passes null to PHP’s current() function. On PHP 8+, this causes a fatal TypeError.

Because this happens during the admin AJAX request, the “Shorten” button displays a spinner indefinitely and the URL is not added. I can shorten any other URL I tried, just not the specific one that's triggering the behavior. It could be a result of a misconfiguration on the external site or maybe some kind of anti-bot behavior. 

### Expected Behavior

Yourls should gracefully handle it when there's no content-type header from an external website.

### Steps to Reproduce

1. Open the YOURLS admin page.
2. Attempt to shorten: https://www.ticketweb.com/event/foreign-dissent-11-wills-pub-tickets/14229244
3. Click Shorten.
4. The button continues spinning and no result is displayed.
5. Reload and attempts to shorten any other URL will be fine, just not the one above.

### Context

The URL I was attempting to shorten is a ticketing website. I don't run the website, but I am affiliated with the show which is why I was trying to shorten it.

In the PHP logs, it fails with this error:
PHP Fatal error: Uncaught TypeError: current(): Argument #1 ($array) must be of type array, null given in includes/functions.php:945

Stack trace:
#0 includes/functions.php(945): current()
#1 includes/functions-shorturls.php(97): yourls_get_remote_title()
#2 admin/admin-ajax.php(21): yourls_add_new_link()
#3 {main}

I'm running PHP 8.2.33-debian.10.260731.1206 on Apache 2.4.38-3+deb10u10.

It looks like the relevant code section is: 
`$_charset` = current( $response->headers->getValues( 'content-type' ) );" in `master/includes/functions.php.`

To work around the problem, I just made a super basic plugin that disables Title fetching. That way I can activate and deactivate it on a site that exhibits the problem without touching the core code.

ozh:
Indeed. Funny that it took all this time before someone would shorten a link on a page that return no content-type :)

bhelmick:
[...]

PR #4147, opened 2026-08-14
Fix content-type edge cases when fetching titles

ozh:
Issue #4143 raised the case where a server would serve page without a content-type - easy fix with a simple `(array)` casting. While I was at it, I digged and found a few other edge cases where our remote title fetching would fail - all now covered in new unit tests.


dgw (approved):
:shipit:

PR #4156, opened 2026-08-29
1.10.6

ozh:
1.10.6 "oops no dev"


ozh:
Not sure why the changelog diff tripped over line endings, I'm editing all files through the web interface on Linux

Release is [drafted](https://github.com/YOURLS/YOURLS/releases). I think it's a good practice to cut and paste the HTML comment with the todo-list, in each future release.

Merging here, then publishing `1.10.6`, then bumping to `1.10.7-dev`, later today if every body is cool with it.

PR #4164, opened 2026-09-11
feat: dark mode for the admin interface

TheInfamousToTo:
## Dark mode for the admin interface

Adds a self-contained plugin, `user/plugins/dark-mode/`, that gives the admin
area a dark theme. No core file is modified — the theme is pure override, so it
can be switched off by deactivating the plugin.

### How it works

- `plugin.php` hooks `html_head` to load `dark.css`. That action fires after
  every core stylesheet, so the overrides win at equal specificity without
  needing `!important`.
- `dark.css` declares one palette as CSS custom properties on `:root` and
  derives every rule from it, all inside `@media (prefers-color-scheme: dark)`,
  so the theme follows the OS setting and the light theme is untouched. A
  comment at the top of the file marks where an explicit toggle would hook in
  if one is wanted later.
- The plugin also filters `stats_line_options`, `stats_pie_options` and
  `stats_countries_map_options` to set the Google chart background to
  `transparent`, so the charts sit on either theme; the chart labels are
  recoloured in CSS.
- `.gitignore` gains `!user/plugins/dark-mode/` alongside the existing plugin
  exceptions.

### Coverage

Surfaces from `style.css`, `infos.css`, `cal.css`, `share.css` and
`tablesorter.css`: page frame and menu, forms and action buttons, the add-URL
box, the delete confirmation dialog, the notify bar, sortable tables and
pagination, the stats tabs and historical clicks list, the date picker, and the
share box.

Colours that carry meaning rather than decoration were kept distinguishable
instead of being inverted: the error, warning and success states each keep
their own hue against the dark background. The historical clicks rows need
`!important` because `js/infos.js` writes their striping as an inline style.

### How to test

Activate **Dark Mode** from Manage Plugins, then switch the OS colour scheme
and compare the admin index, a stats page, the tools page and the plugins page
in both schemes. With the OS in light mode nothing should change at all.

Feedback on the palette is welcome — the values are all in the `:root` block at
the top of `dark.css`, so they are easy to adjust in one place.

---

- [x] I have read the Contributing Guidelines
- [x] I am using AI and have read the AI Policy

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,6 +18,7 @@ user/*
 !user/plugins/sample-page/
 !user/plugins/sample-plugin/
 !user/plugins/sample-toolbar/
+!user/plugins/dark-mode/
 # ... example pages
 !user/pages/examplepage.php
 # ... and index.html files
--- a/user/plugins/dark-mode/dark.css
+++ b/user/plugins/dark-mode/dark.css
@@ -1,0 +1,293 @@
+/* Dark theme for the YOURLS admin interface.
+ *
+ * One palette, declared once as custom properties; every rule below reads from
+ * it. Wrapped in prefers-color-scheme so it follows the OS setting. To add an
+ * explicit toggle later, repeat the selectors as html[data-theme="dark"] and
+ * move the rules out of the media query.
+ */
+@media (prefers-color-scheme: dark) {
+
+:root {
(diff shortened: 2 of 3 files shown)


ozh:
Please : read the contributing.md, the AI policy, and make a plugin

